PR #14, opened 2016-12-19
Handle Facebook logins without email

### LINKED ISSUE #12: Re-request Facebook authentication dialog if user declines to give her email
If the modifies the Facebook app authentication dialog and declines to give her email address, SSO authentication will fail. When the user tries to log in again, no app authentication dialog is shown by default. helsso needs to specifically request this by passing `auth_type=rerequest` to Facebook.

https://developers.facebook.com/docs/facebook-login/web/permissions#re-asking-declined-permissions

suutari-ai:
If user didn't give a consent for passing e-mail address from Facebook
to Helsso, show a notice that e-mail is really needed and try again.  If
user still doesn't give the e-mail address, redirect to login cancelled
view.

Fixes #12

### DIFF
--- a/helsso/urls.py
+++ b/helsso/urls.py
@@ -25,6 +25,8 @@ def show_login(request):
 urlpatterns = [
     url(r'^admin/', include(admin.site.urls)),
     url(r'^accounts/profile/', show_login),
+    url(r'^accounts/login/', LoginView.as_view()),
+    url(r'^accounts/logout/', LogoutView.as_view()),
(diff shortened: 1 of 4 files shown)


suutari-ai:
Note: Might need some tuning still, since the "canceled" view links to the django-allauth login view, which is not the login view we want.

suutari-ai:
OK, the django-allauth login view is now overridden by our own login view and this PR should be OK.

PR #19, opened 2017-06-08
Use new Helsinki brand. Restyle login and authorize -pages

terotic:
Not sure if settings.py:
```
STATICFILES_DIRS = [
    ('node_modules', 'node_modules/'),
]
NODE_MODULES_URL = STATIC_URL + 'node_modules/'
```
and 
application_variables.scss:
```$icon-font-path: unquote(get-setting(NODE_MODULES_URL) + "bootstrap-sass/assets/fonts/bootstrap/");```

is the best way to make sure fonts are being loaded by sass from npm dependency.

### CODE REVIEW

File: helsinki_theme/static/css/helsinki_theme.scss [hunk 1]
```diff
@@ -0,0 +1,58 @@
...
+
+// Bootstrap with custom variables
+@import 'hel-bootstrap-3/src/sass/theme-bootstrap-variables';
+@import 'bootstrap-sass/assets/stylesheets/bootstrap';
+
+// Helsinki Bootstrap overrides
+@import 'hel-bootstrap-3/src/sass/theme-custom-styles';
+
+// Bootstrap Social
+@import 'bootstrap-social/bootstrap-social.scss';
```
terotic: Needs to be explicitly .scss as the package contains .css file with same name in same directory

juyrjola:
Merged slightly manually.

PR #21, opened 2017-06-28
Prefer POST method for API Tokens request

suutari-ai:
This is more in line with OAuth2.


suutari-ai:
@juyrjola: So, should we do this?

PR #22, opened 2017-06-28
User consent storing of public clients

suutari-ai:
Patch the `get` method of oidc_provider.views.AuthorizeView so that it
allows using stored user consents.

This is needed, because Django OIDC Provider doesn't allow stored User
Consent to be used if client type is public.

This hack replaces the value "public" with another value while doing the
authorize request in django_oidc and finally restores the original
value.

### DIFF
--- a/tunnistamo/oidc.py
+++ b/tunnistamo/oidc.py
@@ -1,3 +1,6 @@
+import oidc_provider.views
+
+
 def sub_generator(user):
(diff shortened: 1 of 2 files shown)


suutari-ai:
Too much a hack. Let's hope that we'll get a new version of django-oidc-provider soon. Until that forking would be better way (see PR #23).

PR #30, opened 2017-09-21
Travis: Run "prequ check"

suutari-ai:
Make sure that the compiled requirements are coherent by running
`prequ check -v` on Python 3.6 builds.  This command will fail, if
requirements.in is updated, but requirements.txt is not or if there is
a requirement conflict (caused by manually editing requirements.txt).


.travis.yml
```diff
@@ -7,6 +7,15 @@ cache: pip
 
 install: 'pip install pytest-cov -r requirements.txt'
 
-script: pytest --cov=.
+script:
+  # Run the tests
+  - pytest --cov=. || true  # TODO. Remove the "|| true" when there is tests
```
hukka: We should just let the travis fail, as without any tests the CI result doesn't really give us any confirmation on quality

vikoivun: Actually, here the knowledge whether requirements are valid or not, is already useful information. We had a real in-production failure because of this.

suutari-ai: Having just a single test doesn't confirm the quality much better. Confirmation of quality should come from the test coverage. I would suggest that you integrate to codecov.io or similar for that. Then you would get the info about (a) how well are the parts changed in the PR covered by the tests and (b) how well is the project tested over all.

juyrjola:
Merged this one manually.

PR #31, opened 2017-10-03
Add ability to choose login methods per OIDC Client

### LINKED ISSUE #32: Add ability to choose login methods per OIDC Client

mikkokeskinen:
By recommendation from Suutari, I added a new OidcClientOptions model which has the same additional fields as the Applications model has.

By default, if there are no options saved for the client, all login methods are available.

Closes #32


users/views.py
```diff
@@ -20,23 +21,41 @@ class LoginView(TemplateView):
...
+
             if client_id:
                 try:
                     app = get_application_model().objects.get(client_id=client_id)
                 except get_application_model().DoesNotExist:
                     pass
+
+                try:
+                    oidc_client = Client.objects.get(client_id=client_id)
+                except get_application_model().DoesNotExist:
```
suutari-ai: This should probably be Client.DoesNotExist

suutari-ai: Though IMHO `oidc_client = Client.objects.filter(client_id=client_id).first()` would be cleaner anyway.

mikkokeskinen: Yeah it should. I'll fix it.

PR #40, opened 2018-02-21
Add after_userlogin_hook

mikkokeskinen:
The hook is used to keep the Django session fresh when the user
authorizes themselves using oidc.


suutari-ai:
Btw, the Travis complaint was

    ./users/views.py:18:5: C901 'LoginView.get' is too complex (19)

So completely unrelated to this one. (That method should be refactored though, but in another PR)

PR #63, opened 2018-09-20
OIDC provider clean-up

suutari-ai:
Update django-oidc-provider and get rid of ID token processing hook.
This also removes the ugly hack which used inspect to steal local
variables from the caller (was in id_token.py).

Also remove unneeded FakeToken class and use the original token instead.

See the commit messages for more details.

NOTE: Now there is an unit test for the ID token content, but I still haven tested this in real scenario (i.e. logging in via any real OIDC client).  Though, now I'm quite confident it will work.


users/tests/test_models.py
```diff
@@ -116,6 +119,75 @@ def test_implicit_oidc_login_user_login_entry_creation(client, oidc_client, user
...
+    parsed_loc = urlparse(redirect_location)
+    params = parse_qs(parsed_loc.fragment)
+
+    id_token = params['id_token'][0]
+    id_token_data = jwt.decode(id_token, verify=False)
+
+    expected_keys = {
+        'aud', 'sub', 'exp', 'iat', 'iss',  'nonce',
+        'at_hash', 'auth_time',
+    } | ({
```
tituomin: This expression is a bit "clever", but I'll let it pass. :)
Or maybe it's idiomatic, I don't know.

tituomin: Especially lines 150 and 152

suutari-ai:
>  Do you have an idea how to test this locally "in real life" or whether it's necessary? Maybe just a basic openid connect flow using facebook.

Best test would be to login locally with some app that uses the data from the ID token. The identity provider used for the delegated authentication (Facebook, Google, etc.) shouldn't make a difference, so Facebook is fine. Just check that the app (e.g. helpt) gets the correct data from Tunnistamo: user real name and/or email for example.

suutari-ai:
Is this still waiting for more manual testing or can this be merged already?

Note: This would make it easier to also do further updates to django-oidc-provider in Tunnistamo, since updating it to newer version will require changes like this anyway.

tituomin:
Still waiting. Will increase priority.

PR #73, opened 2018-10-29
Add also scopes required by APIs to UserConsent objects

### LINKED ISSUE #71: Add scopes an API requires to corresponding UserConsent objects

tuomas777:
Previously if there was an API, which required some scope "X", that scope X
was displayed in the consent screen, but it was not added to the UserConsent
object generated as result of the authorization. Therefore it was not
possible to figure out later all the scopes a user has given consent to.

This implementation involves manipulating GET and POST params before the
authorization view, so it is a bit hacky solution, but probably still the
cleanest possible as there is no way to customize Django OIDC Provider's
authorization process.

Closes #71

### DIFF
--- a/oidc_apis/factories.py
+++ b/oidc_apis/factories.py
@@ -1,0 +1,37 @@
+import factory
+
+from users.factories import OIDCClientFactory
+
+from .models import Api, ApiDomain, ApiScope
+
+
+def create_oidc_client_for_api(api):
+    domain_identifier = api.domain.identifier.rstrip('/')
+    return OIDCClientFactory(client_id='{}/{}'.format(domain_identifier, api.name))
+
+
+class ApiDomainFactory(factory.django.DjangoModelFactory):
+    identifier = factory.Faker('url')
+
+    class Meta:
+        model = ApiDomain
+
+
+class ApiFactory(factory.django.DjangoModelFactory):
+    domain = factory.SubFactory(ApiDomainFactory)
+    name = factory.Faker('word')
+    oidc_client = factory.LazyAttribute(create_oidc_client_for_api)
+
+    class Meta:
+        model = Api
+
+
+class ApiScopeFactory(factory.django.DjangoModelFactory):
+    api = factory.SubFactory(ApiFactory)
+    specifier = ''
+    identifier = factory.LazyAttribute(ApiScope._generate_identifier)
+    name = factory.Faker('word')
+    description = factory.Faker('sentence')
+
+    class Meta:
+        model = ApiScope
--- a/oidc_apis/models.py
+++ b/oidc_apis/models.py
@@ -9,6 +9,8 @@ from parler.fields import TranslatedField
 from parler.managers import TranslatableQuerySet
 from parler.models import TranslatableModel, TranslatedFieldsModel
 
+from oidc_apis.utils import combine_uniquely
+
 from .mixins import AutoFilledIdentifier, ImmutableFields
 
 alphanumeric_validator = RegexValidator(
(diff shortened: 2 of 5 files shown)


oidc_apis/factories.py
```diff
@@ -0,0 +1,33 @@
...
+
+    class Meta:
+        model = ApiDomain
+
+
+class ApiFactory(factory.django.DjangoModelFactory):
+    domain = factory.SubFactory(ApiDomainFactory)
+    name = factory.Faker('word')
+    oidc_client = factory.LazyAttribute(
+        lambda o: OIDCClientFactory(client_id='{}/{}'.format(o.domain.identifier.rstrip('/'), o.name)))
```
suutari-ai: a free function would make this clearer than a lambda

oidc_apis/factories.py
```diff
@@ -0,0 +1,33 @@
...
+
+    class Meta:
+        model = Api
+
+
+class ApiScopeFactory(factory.django.DjangoModelFactory):
+    api = factory.SubFactory(ApiFactory)
+    specifier = ''
+    identifier = factory.LazyAttribute(lambda o: ApiScope._generate_identifier(o))
```
suutari-ai: isn't this same as `factory.LazyAttribute(ApiScope._generate_identifier)`?

oidc_apis/models.py
```diff
@@ -186,3 +188,15 @@ class Meta:
 
     def __str__(self):
         return "{obj}[{lang}]".format(obj=self.master, lang=self.language_code)
+
+
+def get_all_required_scopes_by_api_scopes(scopes):
```
suutari-ai: I don't think this needs to be public. It is also very much related to ApiScope class, so maybe it should be a classmethod there?

oidc_apis/models.py
```diff
@@ -186,3 +188,15 @@ class Meta:
...
+    api_scopes = ApiScope.objects.by_identifiers(scopes)
+    apis = {x.api for x in api_scopes}
+    return set(sum((list(api.required_scopes) for api in apis), []))
+
+
+def extend_scope_by_api_scopes(scopes):
```
suutari-ai: How about moving this to a class method such as `ApiScope.extend_scope` (shorter method name should be enough, since the ApiScope class in the name should make it clear that it's somehow related to ApiScopes anyway)

oidc_apis/models.py
suutari-ai: e.g. `ApiScope._get_required_scopes`

users/tests/test_tunnistamo_authorize_view.py
```diff
@@ -36,3 +38,50 @@ def test_tunnistamo_authorize_view_language(client):
...
+        'client_id': oidc_client.client_id,
+        'redirect_uri': oidc_client.redirect_uris[0],
+        'scope': '{} github_username'.format(api_scope.identifier) if api_scope_in_request else api_scope.identifier,
+        'response_type': 'code',
+        'allow': True,
+    })
+    assert response.status_code == 302
+
+    user_consent = UserConsent.objects.get(user=user, client=oidc_client)
+    assert 'github_username' in user_consent.scope
```
suutari-ai: Good test!

users/views.py
```diff
@@ -116,8 +118,29 @@ def get_context_data(self, **kwargs):
...
                 return super().get(request, *args, **kwargs)
+
         return super().get(request, *args, **kwargs)
+
+    def post(self, request, *args, **kwargs):
+        scope = request.POST.get('scope')
+        if scope:
+            query_params = request.POST.copy()
+            query_params['scope'] = self._add_api_scopes(scope)
+            request.POST = query_params
```
suutari-ai: Could be DRYer, e.g.
```python
class TunnistamoOidcAuthorizeView(AuthorizeView): 
    def get(...):
        request.GET = _extend_scope_in_query_params(request.GET)
        ...

    def post(...):
        request.POST = _extend_scope_in_query_params(request.POST)
        ...


def _extend_scope_in_query_params(query_params):
    scope = query_params.get('scope')
    if scope:
        query_params = query_params.copy()
        query_params['scope'] = _add_api_scopes(scope)
    return query_params
```

Note: assumed _add_api_scopes to be moved outside of the class too. (I think free methods usually make more sense than staticmethods)

suutari-ai:
Nice work! Suggested some minor changes though, see the comments above.

tituomin:
Fine by me, if suutari is happy.

tuomas777:
Nice comments, implemented all suggested changes.

PR #74, opened 2018-11-20
Add API list endpoint for OIDC scopes

tuomas777:
The endpoint includes all the scopes: the default ones coming from Django
OIDC provider, the custom ones defined in CombinedScopeClaims class, and
also the scopes for the APIs.

Refs https://github.com/City-of-Helsinki/omahelsinki/issues/136

### DIFF
--- a/scopes/api.py
+++ b/scopes/api.py
@@ -1,0 +1,90 @@
+from django.conf import settings
+from django.utils import translation
+from django.utils.translation.trans_real import translation as trans_real_translation
+from rest_framework import serializers
+from rest_framework.schemas import AutoSchema
+from rest_framework.views import APIView
+
+from oidc_apis.models import ApiScope
+from oidc_apis.scopes import CombinedScopeClaims
+from tunnistamo.pagination import DefaultPagination
+from tunnistamo.utils import TranslatableSerializer
+
+ENGLISH_LANGUAGE_CODE = 'en'
+LANGUAGE_CODES = [l[0] for l in settings.LANGUAGES]
+assert ENGLISH_LANGUAGE_CODE in LANGUAGE_CODES
+
+
+class ApiScopeSerializer(TranslatableSerializer):
+    id = serializers.CharField(source='identifier')
+
+    class Meta:
+        model = ApiScope
+        fields = ('id', 'name', 'description')
+
+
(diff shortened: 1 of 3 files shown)


tunnistamo/urls.py
```diff
@@ -51,7 +52,8 @@ def get_schema(self, *args, **kwargs):
 router.register('service', ServiceViewSet)
 router.register('user_consent', UserConsentViewSet)
 
-v1_api_path = path('v1/', include((router.urls, 'v1')))
+v1_scope_path = path('scope/', ScopeListView.as_view(), name='scope-list')
```
tituomin: Why is this not under v1?

scopes/api.py
```diff
@@ -0,0 +1,90 @@
...
+from rest_framework.views import APIView
+
+from oidc_apis.models import ApiScope
+from oidc_apis.scopes import CombinedScopeClaims
+from tunnistamo.pagination import DefaultPagination
+from tunnistamo.utils import TranslatableSerializer
+
+ENGLISH_LANGUAGE_CODE = 'en'
+LANGUAGE_CODES = [l[0] for l in settings.LANGUAGES]
+assert ENGLISH_LANGUAGE_CODE in LANGUAGE_CODES
```
tituomin: Importing will fail if english is not supported?

tunnistamo/urls.py
juyrjola: Indeed. You can conveniently include API endpoints from different Django apps under the same router like this: https://github.com/City-of-Helsinki/aplans/blob/master/aplans/urls.py#L26

scopes/api.py
tuomas777: Well, the endpoint won't work if english isn't supported, so I figured it would be better to fail soon. How would you like it to work?

tunnistamo/urls.py
tuomas777: > Why is this not under v1?

It is under v1, see next line `v1_api_path = path('v1/', include((router.urls + [v1_scope_path], 'v1')))`.

> Indeed. You can conveniently include API endpoints from different Django apps under the same router like this: https://github.com/City-of-Helsinki/aplans/blob/master/aplans/urls.py#L26

Sure, but you can only register a `ViewSet` to a router, not an `ApiView` which is the case here.

PR #75, opened 2018-12-07
API: Allow nesting scope data in user consent data

tuomas777:
Implemented ability to include full scope data instead of just ids by providing
query parameter "include=scope" in the user consent endpoint request.

### DIFF
--- a/scopes/api.py
+++ b/scopes/api.py
@@ -1,5 +1,6 @@
 from django.conf import settings
 from django.utils import translation
+from django.utils.functional import cached_property
 from django.utils.translation.trans_real import translation as trans_real_translation
 from rest_framework import serializers
 from rest_framework.schemas import AutoSchema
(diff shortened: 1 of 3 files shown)


scopes/api.py
```diff
@@ -52,11 +48,36 @@ def paginator(self):
...
-    def _create_oidc_scopes_data(cls):
-        if hasattr(cls, '_oidc_scopes_data'):
-            return cls._oidc_scopes_data
 
+class ScopeDataBuilder:
+    """
+    A builder for scope data to be used in the API.
+
+    Implemented as a class to provide better control of caching. A ScopeDataBuilder instance caches
+    ApiScopes (which are in the db), so in practice there probably should be an own instance per request
```
tituomin: "in practice there probably should be" ... Can you make this a bit more specific?

PR #76, opened 2019-01-04
Reduce amount of personal data in profile scope

Rikuoja:
Currently, the standard profile scope is requested by all client applications.

"Basic" information should definitely not contain sensitive personal data such as gender and birth date, as they are not required by any of our client applications, and we do not want them to be saved in all user databases.

Also, we do not want to ask the user for permission to give such personal data to all client applications, as current Tunnistamo implementation does; the user should know exactly what data they give permission to, and should not be asked for any data that is not needed.

Currently, the Tunnistamo db does not contain such data, but at some point in the future (e.g. suomi.fi integrations), it might come to pass that such information is available, and it should be in a strictly different scope if ever needed.

Certainly, the oidc standard lists those fields as being part of "basic" profile information, but luckily implementations such as that of OAuth (https://auth0.com/docs/scopes/current/oidc-scopes) have omitted birthdate and gender.


oidc_apis/scopes.py
```diff
@@ -85,7 +85,20 @@ class CustomInfoTextStandardScopeClaims(StandardScopeClaims):
...
     def scope_profile(self):
-        return super().scope_profile()
+        dic = {
+            'name': self.userinfo.get('name'),
+            'given_name': (self.userinfo.get('given_name') or
+                           getattr(self.user, 'first_name', None)),
+            'family_name': (self.userinfo.get('family_name') or
+                            getattr(self.user, 'last_name', None)),
+            'middle_name': self.userinfo.get('middle_name'),
+            'nickname': self.userinfo.get('nickname') or getattr(self.user, 'username', None),
```
suutari-ai: Is it a good idea to pass username as a nickname? In my opinion, it's not and IIRC there's already some code in Tunnistamo to override that behavior of django-oidc-provider.

suutari-ai: See this comment and code:

https://github.com/City-of-Helsinki/tunnistamo/blob/96fe16a9ed012c43c73ea42e4898aedbc13fb857/tunnistamo/oidc.py#L32-L41

suutari-ai:
This is a very good change! :+1: 

But still, see my comment about the username. I might be better to implement it more in line with other Tunnistamo code.

oidc_apis/scopes.py
tituomin: @Rikuoja : @suutari-ai has a point. Otherwise, I'm ready to merge this.

oidc_apis/scopes.py
```diff
@@ -66,10 +66,10 @@ def scope_ad_groups(self):
         }
 
 
-class CustomInfoTextStandardScopeClaims(StandardScopeClaims):
+class ReducedStandardScopeClaims(StandardScopeClaims):
     info_profile = (
         _('Basic profile'),
-        _('Access to your basic information. Includes names, gender, birthdate and other information.'),
+        _('Access to your basic information. Includes names, nickname and possible user picture.'),
```
OP-Lamminen: This message could only mention "names and a possible user picture" as the nickname in userinfo is set as the user's first name in `tunnistamo.oidc.get_userinfo` in the line below (and nickname could be considered to be included in 'names').
https://github.com/City-of-Helsinki/tunnistamo/blob/c4d622f29f6bfaf0269f30892415eabd7226d8c6/tunnistamo/oidc.py#L41

oidc_apis/scopes.py
tituomin: @suutari-ai actually, the nickname in ```self.userinfo.get('nickname')``` has already been initialized by calling the code you referenced above, so the or-part is never called. However, to be explicit, we might remove the or part from this code too.

PR #82, opened 2019-01-21
Feature/rsa key rotation

### LINKED ISSUE #81: Rotate RSA keys at set intervals
Tunnistamo uses RSA keys to sign/encrypt ID tokens. These keys should be rotated at set intervals.

OP-Lamminen:
Fixes #81


key_manager/management/commands/manage_openid_keys.py
```diff
@@ -0,0 +1,108 @@
...
+        """
+        try:
+            key = RSA.generate(length)
+            rsakey = RSAKey(key=key.exportKey('PEM').decode('utf8'))
+            rsakey.save()
+            self.stdout.write('Created new key of length {0} with id: {1}'.format(length, rsakey))
+            return rsakey
+        except Exception as e:
+            self.stdout.write('Something went wrong: {0}'.format(e))
+            raise CommandError('Could not create RSA key: {0}'.format(e))
```
suutari-ai: I think it would be better to not catch exceptions that are not really handled. There would be better debugging info available (e.g. traceback) if this wasn't catched.

key_manager/management/commands/manage_openid_keys.py
```diff
@@ -0,0 +1,108 @@
...
+        """
+        Start managing an unmanaged RSA key.
+        """
+        managedkey = None
+        try:
+            managedkey = ManagedRsaKey.objects.get(pk=rsakey)
+        except ObjectDoesNotExist:
+            managedkey = ManagedRsaKey(key_id=rsakey, created=date.today())
+            managedkey.save()
+        return managedkey
```
suutari-ai: This can be done easier with `get_or_create`
```python
return ManagedRsaKey.objects.get_or_create(
    key_id=rsakey,
    defaults={'created': date.today()},
)[0]
```

Btw. wouldn't it be nicer to have the `created` field be datetime?

key_manager/management/commands/manage_openid_keys.py
```diff
@@ -0,0 +1,81 @@
...
+        if options['list_only']:
+            return
+
+        valid_keys = False
+        # loop through all the keys
+        for rsakey in RSAKey.objects.all():
+            # check if key is managed, and if not, create it
+            managedkey, created = ManagedRsaKey.objects.get_or_create(
+                key_id=rsakey,
+                defaults={'created': timezone.now(), 'expired': timezone.now()})
```
juyrjola: Instead of repeatedly calling `timezone.now()`, you might want to do `now = timezone.now()` before the loop.

PR #83, opened 2019-01-23
Add a .well-known URL to the root of the instance

juyrjola:
The well-known URI should be placed in the root of the instance for discovery mechanism to work properly. This PR adds another route to the `oidc_provider` view `ProviderInfoView`. The old URL still works normally for backwards compatibility.

### DIFF
--- a/tunnistamo/urls.py
+++ b/tunnistamo/urls.py
@@ -10,6 +10,7 @@ from django.views.defaults import permission_denied
 from rest_framework.documentation import include_docs_urls
 from rest_framework.routers import SimpleRouter
 from rest_framework.schemas import SchemaGenerator
+from oidc_provider.views import ProviderInfoView as OIDCProviderInfoView
 
 from devices.api import UserDeviceViewSet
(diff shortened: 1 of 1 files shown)


suutari-ai:
Does this make any difference if the Django app isn't served from the root of the domain (isn't it currently served from `/sso`)?

juyrjola:
It shouldn't. In that case, the issuer URI should be `https://api.hel.fi/sso`, which is allowed in the OIDC specs. Currently, however, the issuer URI is formed like this:

```
def get_issuer(site_url=None, request=None):
    """
    Construct the issuer full url. Basically is the site url with some path
    appended.
    """
    site_url = get_site_url(site_url=site_url, request=request)
    path = reverse('oidc_provider:provider-info') \
        .split('/.well-known/openid-configuration')[0]
    issuer = site_url + path

    return str(issuer)
```

This means the issuer URI will still contain the `/openid` path for the OpenID implementation, which is not desirable. A good time to fix that would be when the Helsinki Tunnistamo instance is moved to another location.

PR #84, opened 2019-03-08
Suomi.fi authentication

OP-Lamminen:


tunnistamo/tests/test_suomifi_authentication.py
```diff
@@ -0,0 +1,353 @@
...
+    expected_metadata = load_file('suomifi_metadata.xml')
+    assert metadata_response.status_code == 200
+    assert metadata_response.get('content-type') == 'text/xml'
+    assert metadata_response.content == expected_metadata
+
+
+@pytest.mark.django_db
+@freeze_time('2019-01-01 12:00:00', tz_offset=2)
+def test_suomifi_login_request(django_client, fixed_saml_id):
+    '''Suomi.fi use case #1'''
```
tituomin: Where are these use cases documented and numbered? I'll store the reference in an appropriate place.

tunnistamo/settings.py
```diff
@@ -383,9 +387,123 @@
 SOCIAL_AUTH_ESPOO_ADFS_KEY = ''
 SOCIAL_AUTH_ESPOO_ADFS_SECRET = None
 
-IPWARE_META_PRECEDENCE_ORDER = ('REMOTE_ADDR',)
+# Social Auth Suomi.fi values. These should be overwritten in local settings.
```
tituomin: Should *all* of these be overwritten? For example SOCIAL_AUTH_SUOMIFI_SP_EXTRA ?
Do the test settings contain enough of a documentation (in the form of an example) of what should be configured and how, or would we need a separate documentation?

tunnistamo/tests/test_suomifi_authentication.py
OP-Lamminen: https://palveluhallinta.suomi.fi/fi/sivut/tunnistus/kayttoonotto/asiakastestiymparisto

tunnistamo/settings.py
OP-Lamminen: Settings have been reordered and comments are now a bit more verbose.

PR #85, opened 2019-03-26
Suomi.fi access levels and claims

OP-Lamminen:
Configurable scopes and claims for Suomi.fi authentication provider.
- Configuration of access levels and attributes is defined in a YAML file
- Requesting a Suomi.fi scope not listed in OIDC clients configuration results in 403-insufficient_scope error


oidc_apis/scopes.py
```diff
@@ -97,6 +101,44 @@ def scope_address(self):
...
+    def __dir__(cls):
+        names = super().__dir__()
+        for level in SuomiFiAccessLevel.objects.all():
+            names.append('info_suomifi_' + level.shorthand)
+        return names
+
+    def __getattr__(cls, name):
+        if name.startswith('info_suomifi_'):
+            try:
+                level = SuomiFiAccessLevel.objects.get(shorthand=name[13:])
```
tituomin: I understood where this 13 comes from, but still...
How about

  SUOMIFI_ATTR_PREFIX = re.compile('^info_suomifi_')

and then

  shorthand=re.sub(SUOMIFI_ATTR_PREFIX, '', name)

Or something like that?

suutari-ai: Or use the regular expression for both, matching the prefix and picking the correct substring:

```python
        match = re.match(r'^info_suomifi_(.*)', name)
        if match:
            try:
                level = SuomiFiAccessLevel.objects.get(shorthand=match.group(1))
```

OP-Lamminen: Updated.

oidc_apis/scopes.py
```diff
@@ -97,6 +103,45 @@ def scope_address(self):
...
+        for level in SuomiFiAccessLevel.objects.all():
+            names.append('info_suomifi_' + level.shorthand)
+        return names
+
+    def __getattr__(cls, name):
+        match = re.match(r'^info_suomifi_(.*)', name)
+        if match:
+            try:
+                level = SuomiFiAccessLevel.objects.get(shorthand=match.group(1))
+                return (_(level.name), _(level.description))
```
suutari-ai: Using `ugettext` or `ugettext_lazy` (the `_` here) doesn't work with dynamic content, since the argument to ugettext must be a static string literal so that it can be picked up for translation. You should rather use some statically translated name in the model like an enumeration or alternatively provide translated fields in the model, e.g. with Django Parler (check e.g. `services/models.py` for an example).

OP-Lamminen: Added translations to model.

auth_backends/suomifi.py
```diff
@@ -281,8 +287,16 @@ def extra_data(self, user, uid, response, details=None, *args, **kwargs):
...
         data = super().extra_data(user, uid, response, details=details, *args, **kwargs)
         data['session_index'] = response.get('session_index')
+        data['suomifi_attributes'] = self._extract_suomifi_attributes(response)
         return data
 
+    def _extract_suomifi_attributes(self, response):
+        attributes = {}
+        for attribute in SuomiFiUserAttribute.objects.all():
+            if attribute.uri in response.get('attributes'):
+                attributes[attribute.friendly_name] = response.get('attributes')[attribute.uri][0]
```
tituomin: The attribute value is a list because the original XML element can contain many children. Is it specified somewhere that there will only be one value?

OP-Lamminen: Although the SAML specification allows for multiple values per attribute the Suomi.fi documentation describing the attributes implies only a single value for each attribute.

PR #92, opened 2019-05-29
Upgrade django-oidc-provider from 0.6.2 to 0.7

tituomin:
- fix response type model change related errors
- rename EndSessionView method "get" to "dispatch"


users/tests/conftest.py
```diff
@@ -78,11 +78,14 @@ def make_instance(**args):
         args.setdefault('name', get_random_string())
         args.setdefault('client_type', 'public')
         args.setdefault('client_id', get_random_string())
-        args.setdefault('response_type', 'id_token token')
         args.setdefault('redirect_uris', None)
 
+        response_types = args.pop('response_types', ['id_token token'])
```
suutari-ai: Should it be `['id_token', 'token']` instead (a list with two items rather than one)?

users/tests/test_models.py
```diff
@@ -38,7 +38,7 @@ def application():
 
 @pytest.fixture
 def oidc_client():
-    return OIDCClientFactory(response_type='id_token token')
+    return OIDCClientFactory(response_types=['id_token token'])
```
suutari-ai: Same here: Should this be a list of two strings rather than just one?

users/views.py
```diff
@@ -163,7 +163,7 @@ def get(self, request, *args, **kwargs):
             except UserSocialAuth.DoesNotExist:
                 pass
         # clear Django session and get redirect URL
-        response = super().get(request, *args, **kwargs)
+        response = super().dispatch(request, *args, **kwargs)
```
suutari-ai: Commit message says that this is renamed, but doesn't say why. Would help in the review if the reasoning would be available.

users/tests/conftest.py
tituomin: @suutari-ai no, oidc-provider began supporting multiple response types per client in 0.7, and each response type is identified by a string which comes straight from the oidc standard.

We are not using the multiple response types feature in our code or tests, it's just that the ORM level API has changed, and this fixture now supports multiple response types if we need them.

users/tests/test_models.py
tituomin: Same answer.

users/tests/conftest.py
suutari-ai: Ah, then this is OK.

users/tests/test_models.py
suutari-ai: ...and this.

users/views.py
tituomin: @suutari-ai I'll amend the commit. The reason is that django-oidc-provider changed the view to be a base class of LogoutVIew support Django 2.1 (see https://github.com/juanifioren/django-oidc-provider/commit/972071e370cdc0d2798c524e6a4b09681cc727af ) and that resulted in our custom subclass not being called.

tituomin: And that was actually, because in the URL setup, the as_view -method results in dispatch being called. Previously, it was the dispatch of a base view higher up in the class hierarchy, but now it was resolved directly to django-oidc-provider's dispatch which overrided any default dispatch higher up that would have called our get-method.

suutari-ai: OK, thanks for explaining. Seems very clear now. :+1:

PR #96, opened 2019-11-01
Implement new logout screen

tituomin:
Features: warning about lingering 3rd party IDP sessions.

Internationalized template and warnings.

Implement redirection from old logout screens to new,
with query parameter translation.

Adjust tests so that they no longer expect post logout
HTTP redirects, but expect a 200 response with HTML with
a link to the post logout uri.

### DIFF
--- a/auth_backends/github.py
+++ b/auth_backends/github.py
@@ -1,0 +1,21 @@
+from django.utils.translation import gettext as _
+from social_core.backends.github import GithubOAuth2
+
+
+class Github(GithubOAuth2):
+    persistent_session_warning = _(
+        "Please note that you are still logged in to GitHub"
+    )
+    persistent_session_suggestion = _(
+        "Go back to GitHub where you can log yourself out "
+        "in the usual manner:"
+    )
+    persistent_session_link = _(
+        "Go to GitHub"
+    )
+    persistent_session_final_warning = _(
+        "If access to this device is shared by other users, "
+        "the next user will be able to access your GitHub account "
+        "unless you explicitly log out from GitHub."
+    )
+    user_facing_url = 'https://github.com'
--- a/auth_backends/github.py
+++ b/auth_backends/github.py
@@ -1,21 +1,12 @@
-from django.utils.translation import gettext as _
+from django.utils.translation import gettext, pgettext
 from social_core.backends.github import GithubOAuth2
 
 
 class Github(GithubOAuth2):
-    persistent_session_warning = _(
-        "Please note that you are still logged in to GitHub"
-    )
-    persistent_session_suggestion = _(
-        "Go back to GitHub where you can log yourself out "
-        "in the usual manner:"
-    )
-    persistent_session_link = _(
-        "Go to GitHub"
-    )
-    persistent_session_final_warning = _(
-        "If access to this device is shared by other users, "
-        "the next user will be able to access your GitHub account "
-        "unless you explicitly log out from GitHub."
-    )
     user_facing_url = 'https://github.com'
+    name_baseform = gettext('GitHub')
+    name_access = pgettext('access to []', 'GitHub')
+    name_genetive = pgettext('genetive form', 'GitHub')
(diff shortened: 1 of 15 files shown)


auth_backends/github.py
```diff
@@ -0,0 +1,21 @@
+from django.utils.translation import gettext as _
+from social_core.backends.github import GithubOAuth2
+
+
+class Github(GithubOAuth2):
+    persistent_session_warning = _(
```
juyrjola: These are starting to violate the DRY principle. Maybe have a mixin class that has the translation strings with the user-readable backend name as a variable?

tituomin: This is by decision.

There is no way extract the backend name as variable without making the Finnish clunky, by which I mean without using language like "Facebook-palveluun, Facebook-palvelusta" etc.

juyrjola: Yes, way! You could use [contextual markers](https://docs.djangoproject.com/en/2.2/topics/i18n/translation/#contextual-markers) to specify the noun case.

tituomin: Alright, alright. I will try.

suutari-ai: I don't think you even need contextual markers for this. Just translate them separately, i.e. something like this:
```python
class Blaa(...):
    @property
    def persistent_session_warning(self):
        return _(
            "Please note that you are still logged in {to_service}"
        ).format(to_service=self.to_service)

    @property
    def persistent_session_link(self):
        return _("Go {to_service}").format(self.to_service)

    @property
    def persistent_session_final_warning(self):
        return _(
            "If access to this device is shared by other users, "
            "the next user will be able to access {your_account} "
            "unless you explicitly log out {from_service}."
        ).format(your_account=self.your_account, from_service=self.from_service)

class GitHubBlaa(Blaa):
    to_service = _("to GitHub")   # GitHubiin
    from_service = _("from GitHub")  # GitHubista
    your_account = _("your GitHub account")  # GitHub-tilillesi
```

But as you can see from the last Finnish translation "GitHub-tilillesi", this form will still make those strings depend on the form used when translating the phrases in the base class (e.g. in this case "toinen käyttäjä voi päästä {your_account}")

suutari-ai: My method works with English, Finnish and Swedish, but might make it impossible to generate good messages for some other languages. In that way @tituomin's original implementation was more predictable.

tituomin: @juyrjola @suutari-ai Please see my new implementation attached to this branch as a commit. https://github.com/City-of-Helsinki/tunnistamo/pull/96/commits/2e2c85c49eceaac36512fa7eb2fe96e81f2354cc

It is true that some DRY can be achieved this way. However, I strongly suspect idiomatic translations sometimes require repetition in translation strings. I can imagine linguists and translators cringing when seeing programmers trying to decompose idiomatic sentences in one language to supposedly language-independent atoms. I myself would have liked to reduce some of the backend-name repetition in these user facing messages, but doing so would have required a mechanism to ensure that pronouns in subordinate clauses would conform to their antedecent's form in the main clause -- doable? Maybe.

I suppose next you will ask me to reduce the repetition of 'Google', 'Google', 'Google' by using intricate trickstery with metaclasses, noop gettext helpers, and lazy translations. Couldn't figure out a way to do that and still have functional makemessages extraction for all the different forms.

suutari-ai: I think the solution implemented in 2e2c85c is beautiful. It's certainly more DRY than the previous one, but also more complex. From programmers perspective the new solution looks nicer. However, from translators perspective, probably the original solution was easier to translate though: it had longer messages and required more typing, but I can imagine that it will take quite some time to figure out what kind of forms should be put into these different contextual forms of Google, Facebook, GitHub and Yle Tunnus. Or maybe I'm wrong and even translators would prefer the second form, since I'm not much of a translator my self.

Nice work @tituomin! :+1: I would approve both solutions. :smiley:

tituomin: @juyrjola want to comment?

PR #98, opened 2019-11-06
Fix CORS tests for openid configuration URL

vikoivun:
Tests /.well-known/openid-configuration URLS ended with slash. That is incorrect:
https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationRequest


juyrjola:
This would need a fix to the `settings.CORS_URLS_REGEX` to make the tests work.

tituomin:
> Fix the CORS regex pattern so that tests will pass.

Done

PR #107, opened 2020-01-22
Security upgrades for Django, Pillow and ecdsa

vikoivun:
Upgrade these to clear the requirements.io alert. Thus the alert will again become useful for detecting any new problems.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,7 +12,7 @@ coreapi==2.3.3
 coreschema==0.0.4         # via coreapi
 cryptography==2.3.1
 defusedxml==0.5.0         # via python3-openid, python3-saml, social-auth-core
-django==2.1.11
+django==2.1.15
 django-allauth==0.36.0
 django-appconf==1.0.2     # via django-compressor
 django-bootstrap3==10.0.1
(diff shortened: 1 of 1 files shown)


vikoivun:
Pillow upgrade has potential to break things.

vikoivun:
> I wasn't able to find any `PIL` imports from the code base so the Pillow dependency is probably used only indirectly through Django. It was added on commit [a341eab](https://github.com/City-of-Helsinki/tunnistamo/commit/a341eab558fc782f923bfee345af203ebdaedac6), which also added an `ImageField`. If this is the only use for it, it should be relatively easy to verify that it still works, e.g. by adding a new Service object with an image from Django Admin and verifying it works correctly.

Indeed. However it looks like we've never even used that functionality. Thus the verifying part is a wee bit tricky. At least I'm still able to upload a file and get the link back from the API

suutari-ai:
> Indeed. However it looks like we've never even used that functionality. Thus the verifying part is a wee bit tricky. At least I'm still able to upload a file and get the link back from the API

The `ImageField` should check that the uploaded content is actually an image. Try uploading garbage to it and it should detect it (raise ValidationError). If that works, then Pillow is doing its job there.

vikoivun:
> > Indeed. However it looks like we've never even used that functionality. Thus the verifying part is a wee bit tricky. At least I'm still able to upload a file and get the link back from the API
> 
> The `ImageField` should check that the uploaded content is actually an image. Try uploading garbage to it and it should detect it (raise ValidationError). If that works, then Pillow is doing its job there.

Thanks for the tip! Uploading a garbage image indeed failed. So Pillow is doing its basic work.

PR #111, opened 2020-04-01
Only allow embedding the authorize endpoint in IFRAME by default

vikoivun:
Previous commit (cbbd1d19ee2f0326c1e1289c65b8b5594d6ed3bd ) changed the default to allow embedding the whole application in IFRAME. Silent renew only requires access to the authorize-endpoint, which is now always allowed to be framed. The switch to allow embedding the
whole application is still present, but defaults to False.

### DIFF
--- a/tunnistamo/settings.py
+++ b/tunnistamo/settings.py
@@ -20,7 +20,7 @@ env = environ.Env(
     SECRET_KEY=(str, ""),
     DATABASE_URL=(str, ""),
     ALLOWED_HOSTS=(list, []),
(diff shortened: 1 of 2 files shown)


vikoivun:
> Is there an actual use case which requires the other views to be available inside an iframe? If not, I would just get rid of the possibility altogether, because of YAGNI and the fact that it can be a potential security issue when misused.

I wish I knew. I don't think so though.

Mostly I'm worried that there might be some other view needed for the silent renew to work. Seems unlikely though.

vikoivun:
Closing. Iframe embedding not needed in the end.

PR #113, opened 2020-04-07
Add environment based config for username / OIDC backend

vikoivun:


vikoivun:
Only added config variables, following existing format. Committed directly.

PR #119, opened 2020-11-04
Remove mentions of prequ, replace with pip-tools, upgrade Django

tituomin:


suutari-ai (approved):
This is probably a good idea, since pip-tools is well maintained these days, but Prequ hasn't got any updates for a long time and doesn't work with new pip anymore.

suutari-ai:
Btw. you should probably also check the headers in the requirements*.txt files, since they contain instruction for running `prequ update`.

README.md
```diff
@@ -100,15 +100,13 @@ To access the themed views you also need to install
...
 to manage the Python dependencies.
-prequ can handle `-e` style dependencies (git URLs) in the requirements files.
 
 Update the requirements with:
 ```
-pip install prequ
-rm requirements.txt
-prequ update
+pip install pip-tools
+pip-compile --upgrade
```
suutari-ai, line 109: This will only handle the main requirements file, so maybe you should mention how to update the dev and prod requirement files too?

PR #127, opened 2020-12-22
Fix base template reference in email verification template

tituomin:
django-allauth 0.43.0 renamed a template to use a name
already used by the signup closed template. This is fixed
by using the correct reference in our template.

https://github.com/pennersr/django-allauth/blob/master/ChangeLog.rst


tituomin:
This fixes a bug which caused a `django.urls.exceptions.NoReverseMatch: Reverse for 'account_email' not found. 'account_email' is not a valid view function or pattern name.`

PR #131, opened 2021-01-08
Upgrade pillow and py, move to pip-tools 5.5.0

tituomin:


requirements-dev.txt
```diff
@@ -1,8 +1,12 @@
...
-# To update, run:
+# NOTE: for the time being, requirements-dev.txt
+# cannot be stably compiled with pip-compile
+# due to an inconsistent pinned version of the
+# six package.
 #
-#    pip-compile requirements-dev.in
+# First upgrade all outdated dependencies in
+# requirements.txt and then compile dev deps
+# again
```
akikoskinen, line 9: Would be nice if this problem with six could be fixed first and then this note here wouldn't be necessary and dependency management could be used normally.

akikoskinen (changes requested):

requirements-dev.txt
tituomin: Ah, this was based on my wrong usage. Removed, fixed.

akikoskinen (approved):
👍

PR #135, opened 2021-01-19
Refactor Dockerfile for Openshift compatibility

vikoivun:
The changes in nutshell:
* everything uid/gid by root (zero) instead of appuser
* group root (zero) is given write access to /var (why?, what is being written runtime)


tituomin (changes requested):
Disable or remove the database check again and let's make it another task & PR altogether.

vikoivun:
Currently it looks like this is not needed. OpenShift has been changed to run builds as root.

tituomin:
Currently it looks like this is not needed. OpenShift has been changed to run builds as root.

PR #136, opened 2021-01-19
Add a toggle for the email duplicate check

vikoivun:
Allows toggling the email duplicate check using ALLOW_DUPLICATE_EMAILS. For backwards compatibility this defaults too false.

Also piggybacked is change that allows the incoming e-mail from the IdP to overwrite the existing e-mail.

### DIFF
--- a/tunnistamo/settings.py
+++ b/tunnistamo/settings.py
@@ -28,6 +28,8 @@ env = environ.Env(
     MEDIA_URL=(str, '/media/'),
     NODE_MODULES_ROOT=(str, os.path.join(BASE_DIR, 'node_modules')),
 
+    ALLOW_DUPLICATE_EMAILS=(bool, False),
+
     # Authentication settings
     SOCIAL_AUTH_FACEBOOK_KEY=(str, ""),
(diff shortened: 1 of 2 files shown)


tunnistamo/settings.py
```diff
@@ -428,7 +434,18 @@
 
     # Save last login backend to user data
     'users.pipeline.save_social_auth_backend'
-)
+]
+
+# We want to have to a positive config option and NOT allowing
+# duplicate emails requires changes. Thus the inverted test.
+if not env("ALLOW_DUPLICATE_EMAILS"):
```
tituomin, line 441: I'd much rather move this logic out of the settings file. Make a new custom pipeline function which calls the original function unless env("ALLOW_DUPLICATE_EMAILS")

tituomin: Or rather, just move the condition to the existing pipeline function because it's our code anyway. I remembered it being part of library code.

tituomin (changes requested):

tituomin (approved):

PR #139, opened 2021-01-22
Allow certain backends to be exempt from email requirement

tituomin:
Some IdPs have users which might not have an email address,
especially Suomi.fi-tunnistus.


vikoivun (approved):
Looking good.

PR #140, opened 2021-01-22
Show the "profile" debug screen when user logs in through /login

vikoivun:
For some reason tunnistamo has defaulted to non-existent /profile URL for logins without redirect back to application. This just changes it to the default /accounts/login, which is marginally more useful.


mikkokeskinen (approved):
Yes, this has been a broken thing for many years. Good change!

PR #141, opened 2021-01-25
Add Tunnistamo authentication backend

tuomas777:
Added a new authentication backend for authenticating users against another Tunnistamo instance using OIDC. It also provides users' AD groups.


vikoivun (approved):
Looks correct. This is needed specifically for transferring AD-groups between two cascaded tunnistamos. Otherwise plain OIDC backend would have sufficed.

PR #146, opened 2021-02-09
Add azp (Authorized party) and loa (Level of Assurance) claims

mikkokeskinen:
The azp claim is simply the current clients client_id. The loa claim is relayed from the HelsinkiTunnistus backend. (i.e. Helsinki Tunnistus Keycloak)

As the OIDC provider and the social auth backends are not connected, the loa value is saved to the users session in the social auth pipeline when the user returns from the HelsinkiTunnistus backend. Afterwards when creating the OIDC tokens, the session value is read and added as "loa" claim. The default value for the "loa" claim is "low".

Refs HP-437

### DIFF
--- a/oidc_apis/scopes.py
+++ b/oidc_apis/scopes.py
@@ -158,7 +158,21 @@ class SuomiFiUserAttributeScopeClaims(ScopeClaims, metaclass=SuomiFiUserAttribut
 
 class OptionalOpenIDScopeClaims(ScopeClaims):
     def scope_openid(self):
-        return {'amr': self.user.last_login_backend} if self.user.last_login_backend else {}
+        claims = {}
+
+        if self.user.last_login_backend:
+            claims['amr'] = self.user.last_login_backend
+
+        # Add the current client id to the "azp" claim (Authorized party - the party
+        # to which the ID Token was issued).
+        claims['azp'] = self.client.client_id
+
+        # Set "low" as the default "Level of Assurance". This is possibly changed in
+        # the "add_heltunnistussuomifi_loa_claim" id processing hook if the user
+        # authenticated using the heltunnistussuomifi provider.
+        claims['loa'] = 'low'
+
+        return claims
 
 
 class CombinedScopeClaims(ScopeClaims):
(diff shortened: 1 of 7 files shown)


akikoskinen (commented):
I'm not familiar with Tunnistamo code, so I can't judge this very closely. But something caught my eye, when talking about the "loa" claim feature. There seems to be very specific, even implementation detail level, terminology: "helsinki" (or "hel"), "suomifi", "keycloak", "loa". If talking on a bit higher abstraction level, isn't the functionality basically something like "copy a claim, that was received from an IdP, into any generated tokens"? How about doing the implementation on that abstraction level, perhaps making the list of copyable claims configurable, or something like that? That would make it more reusable -- and not so implementation dependent.

Another comment: this PR seems to add two unrelated features, the "azp" claim and the "loa" claim. Could the features be added in two separate commits?

mikkokeskinen:
Good points.

I thought about making these more abstract, but as this is supposed to be used for maximum of couple of months, decided against it. Of course if we think this is a feature we'd like to encourage others to use I can definitely use some more effort to make things configurable and extendable. Maybe @tituomin could chime in?

Also it is essentially a hack to connect social authentication with the oidc provider. As they are not designed to use that way at all. As far as I gather.

I could very well separate the changes to two commits but I thought they are so small and they would be squashed to one commit anyway. At least I thought squashing was the way we are supposed to merge.

akikoskinen:
> I thought about making these more abstract, but as this is supposed to be used for maximum of couple of months, decided against it. Of course if we think this is a feature we'd like to encourage others to use I can definitely use some more effort to make things configurable and extendable. Maybe @tituomin could chime in?

Aah, that's correct. Then probably not worth it to do too fancy.

> I could very well separate the changes to two commits but I thought they are so small and they would be squashed to one commit anyway. At least I thought squashing was the way we are supposed to merge.

I don't know what the merging process has been in Tunnistamo, but I sure hope squashing commits isn't the one 😬

tituomin (approved):

tituomin:
I think it's okay for now to do this the low-abstraction way like Mikko argued.

About squashing: no, we want preserve individual commits since they provide documentation and make changes easier to understand. I have sometimes squashed a PR if it contains a small fix in a commit and then another commit fixing a typo or something, but generally our team has agreed that even that kind of cleaning up would be the PR author's responsibility before merging.

PR #147, opened 2021-02-09
Relay client_id from the OIDC provider to the HelsinkiTunnistus backend

mikkokeskinen:
Save the users client_id in the OIDC authorize view to be added later as an extra parameter when the user moves to authenticate with the HelsinkiTunnistus Keycloak.

Notice that the original_client_id is not validated in any way. (And can't be in Tunnistamo when implemented like this) That could lead the user making a service connection to a service they didn't intend to. That of course would need someone to craft a redirect url for the user and trick them somehow.

Refs HP-342

### DIFF
--- a/users/views.py
+++ b/users/views.py
@@ -123,6 +123,19 @@ class AuthenticationErrorView(TemplateView):
 class TunnistamoOidcAuthorizeView(AuthorizeView):
     def get(self, request, *args, **kwargs):
         request.GET = _extend_scope_in_query_params(request.GET)
+
+        if request.GET.get('client_id'):
+            try:
+                client = Client.objects.get(client_id=request.GET.get('client_id'))
+
+                # Save the client_id to the session to be used in the HelsinkiTunnistus
+                # social auth backend.
+                request.session["oidc_authorize_original_client_id"] = client.client_id
+            except Client.DoesNotExist:
+                # We don't care if the client wasn't found because the client will be
+                # validated again in the parent get method.
+                pass
+
         request_locales = [l.strip() for l in request.GET.get('ui_locales', '').split(' ') if l]
         available_locales = [l[0] for l in settings.LANGUAGES]
 
--- a/auth_backends/helsinki_tunnistus_suomifi.py
+++ b/auth_backends/helsinki_tunnistus_suomifi.py
@@ -16,3 +16,19 @@ class HelsinkiTunnistus(OpenIdConnectAuth):
         # We explicitly want to use the UUID from the keycloak provided
         # token as our own identity
         return self.get_user_id(details, response)
+
+    def auth_extra_arguments(self):
+        extra_arguments = super().auth_extra_arguments()
+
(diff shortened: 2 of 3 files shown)


users/views.py
```diff
@@ -131,6 +131,13 @@ def get(self, request, *args, **kwargs):
                 with translation.override(locale):
                     return super().get(request, *args, **kwargs)
 
+        # Save the client_id to the session to be used in the HelsinkiTunnistus
+        # social auth backend.
+        request.session["oidc_authorize_original_client_id"] = request.GET.get(
```
tituomin, line 136: Validation: this is using a query param as-is. I do imagine the actual client id is validated later, but it's kind of implicit. I suppose it might be difficult to only store this value after validation?

tituomin (changes requested):

users/views.py
mikkokeskinen: Yes it will be validated in the parent get method, but I can of course check it here too.

users/views.py
```diff
@@ -123,6 +123,19 @@ def get_context_data(self, **kwargs):
 class TunnistamoOidcAuthorizeView(AuthorizeView):
     def get(self, request, *args, **kwargs):
         request.GET = _extend_scope_in_query_params(request.GET)
+
+        if request.GET.get('client_id'):
+            try:
+                client = Client.objects.get(client_id=request.GET.get('client_id'))
```
tituomin, line 129: Well, I was too slow to comment. I guess the only point that's important is that validation happens before anything is done with the data. If the validation already happens in the super().get -call, could we just move the line where the id is save to `request.session` after the supe().get -call?

mikkokeskinen: Something like this?

```python
class TunnistamoOidcAuthorizeView(AuthorizeView):
    def get(self, request, *args, **kwargs):
        request.GET = _extend_scope_in_query_params(request.GET)

        request_locales = [l.strip() for l in request.GET.get('ui_locales', '').split(' ') if l]
        available_locales = [l[0] for l in settings.LANGUAGES]
        
        response = None
        for locale in request_locales:
            if locale in available_locales:
                with translation.override(locale):
                    response = super().get(request, *args, **kwargs)
                    break

        if not response:
            response = super().get(request, *args, **kwargs)
        
        if request.GET.get('client_id'):
            # Save the client_id to the session to be used in the HelsinkiTunnistus
            # social auth backend.
            request.session["oidc_authorize_original_client_id"] = request.GET.get('client_id')

        return response
```

Maybe could move that locale thing to its own method...

tituomin: Yep, that looks good!

mikkokeskinen: I just realised that a client error ends up in the response as a rendered html... Would need to detect that somehow. I think the manual client checking is the simplest, all things considered.

tituomin: Oh yes, true. Approved!

tituomin (approved):

PR #151, opened 2021-02-16
Added Azure DevOps pipeline

Sukriva:
This pipeline extends the base pipeline and defines CI and PR triggers in GitHub


vikoivun:
Is there something to prevent the whole pipeline (up to prod!) from running automatically with this? I think the Azure side would need be to changed first, to prevent this from happening. I understand you've been talking about splitting the pipelines for the different environments.
Also, given that this is a public repository, potentially used by other cities in the future: would it be possible to have this file named something like `Helsinki-pipelines.yml`

PR #152, opened 2021-02-18
Redirect log out to OIDC providers end session endpoints

mikkokeskinen:
Adds a feature where the user is redirected to end session views of the OIDC providers they used to log in. The setting `REDIRECT_LOGOUT_TO_END_SESSION` must be set to True in the backend to enable the redirect.

The flow is the following:

1. The user loads [tunnistamo host]/openid/end-session
2. TunnistamoOidcEndSessionView checks every UserSocialAuth the user has.

For every social auth that has this feature enabled and the user hasn't been redirected to yet:

3. The backend name is saved to the session to keep track of the backends the user has beed redirected to
4. The user is redirected to the end session endpoint (The address is read from the well-known oidc configuration) with `post_logout_redirect_uri` pointing back to the address in phase 1.

5. The users Django session is removed

Refs HP-448
Refs HP-432


users/tests/test_logout_view.py
```diff
@@ -90,3 +90,138 @@ def test_logout_redirect_next_authenticated(client, user_factory, application_fa
...
+    assert response.url in expected_response_urls
+    response_urls.append(response.url)
+
+    response = client.get('/openid/end-session', params, follow=False)
+
+    assert response.status_code == 302
+    assert response.url in expected_response_urls
+    response_urls.append(response.url)
+
+    assertCountEqual(response_urls, expected_response_urls)
```
akikoskinen, line 222: This is not a biggy, but if I understand this correctly, the redirects can happen in any order and that's why the responses' urls can't be asserted in the simplest way. If my understanding is right, then how about using sets? So, make `response_urls` and `expected_response_urls` sets, collect response urls into `response_urls` (just like currently), and finally `assert response_urls == expected_response_urls`.

users/views.py
```diff
@@ -211,32 +214,111 @@ def get_active_social_users(self, user):
...
+            if not end_session_endpoint:
+                continue
+
+            if 'oidc_backend_end_session_redirected' not in request.session:
+                request.session['oidc_backend_end_session_redirected'] = defaultdict(
+                    bool
+                )
+
+            if request.session['oidc_backend_end_session_redirected'].get(backend.name):
+                continue
```
akikoskinen, line 255: Not sure how the OIDC configs are cached, but should this check be more upwards, perhaps before the place where the `end_session_endpoint` is fetched?

users/views.py
```diff
@@ -211,32 +214,111 @@ def get_active_social_users(self, user):
...
         user = request.user
-        social_user = self._active_suomi_fi_social_user(user)
+
+        social_users = self.get_active_social_users(user)
+        if len(social_users) > 1:
+            logger.warning(
+                'Multiple active social core backends for user {}'.format(
+                    user.pk
+                )
+            )
```
akikoskinen, line 274: Does this need to be logged (I don't know the logging practices in this project, so it can well be) and is warning the correct level? At least the multiple social users case is now handled nicely, so what's there to report? Or perhaps put level to info or something like that?

akikoskinen (commented):
Looks legit to me. But I don't know the internals of this project, so I would like that perhaps e.g. @tituomin also takes a look.

users/tests/test_logout_view.py
mikkokeskinen: True. Was a quick fix and didn't think about it.

akikoskinen:
> SonarCloud Quality Gate failed.
> 
> [Security Hotspot](https://sonarcloud.io/project/security_hotspots?id=City-of-Helsinki_tunnistamo&pullRequest=152&resolved=false&types=SECURITY_HOTSPOT) [E](https://sonarcloud.io/project/security_hotspots?id=City-of-Helsinki_tunnistamo&pullRequest=152&resolved=false&types=SECURITY_HOTSPOT) [19 Security Hotspots](https://sonarcloud.io/project/security_hotspots?id=City-of-Helsinki_tunnistamo&pullRequest=152&resolved=false&types=SECURITY_HOTSPOT)

Sonar seems to warn about using `http` protocol instead of `https`. But those are URLs in test files and they are not used to make any connections anywhere so it shouldn't be a real concern. But perhaps it could start to be a habit to use `https` also in test files...

users/views.py
mikkokeskinen: They are cached. Didn't want to set the variable if there is no end session endpoint.

users/views.py
mikkokeskinen: Don't know. It was there already. I just moved it.

users/views.py
akikoskinen: Makes sense.

users/views.py
tituomin: Info might be better. The multiple social user case is now handled so that there will be a new user for each backend. Still there might be old users that have ended up with multiple social backends. But still, that will have to be handled once we start migrating, and I guess this is the wrong place to complain about it, it's more like a general database data quality issue. Removing is ok by me.

tituomin (approved):

tituomin:
> > SonarCloud Quality Gate failed.
> > [Security Hotspot](https://sonarcloud.io/project/security_hotspots?id=City-of-Helsinki_tunnistamo&pullRequest=152&resolved=false&types=SECURITY_HOTSPOT) [E](https://sonarcloud.io/project/security_hotspots?id=City-of-Helsinki_tunnistamo&pullRequest=152&resolved=false&types=SECURITY_HOTSPOT) [19 Security Hotspots](https://sonarcloud.io/project/security_hotspots?id=City-of-Helsinki_tunnistamo&pullRequest=152&resolved=false&types=SECURITY_HOTSPOT)
> 
> Sonar seems to warn about using `http` protocol instead of `https`. But those are URLs in test files and they are not used to make any connections anywhere so it shouldn't be a real concern. But perhaps it could start to be a habit to use `https` also in test files...

Yeah, this will be corrected when we get Sonar to understand our project-specific test path settings, which will be soon. So it seems that even Sonar thinks this is not relevant for tests once it understands what the tests are. Willing to approve even with the sonar errors for now.

tituomin (approved):

PR #153, opened 2021-03-02
Add CORS_ALLOW_CREDENTIALS setting

mikkokeskinen:
Needed to allow clients browsers to add session cookie when requesting api tokens. Which in turn is needed to read the loa value from the session when generating the api tokens.


vikoivun (approved):
Yup, that is it.

PR #156, opened 2021-03-05
Get loa claim value from the ID Token instead of the session

mikkokeskinen:
The loa value is read from the session in an idtoken_processing_hook
called from create_id_token but the session is not available when API
tokens are requested.

Refs HP-475
Refs HP-437


tituomin:
We should probably remove this setting if it will not be used? https://github.com/City-of-Helsinki/tunnistamo/commit/e6f7bd9be64b410356910814ca5596063b8da4f0

tituomin (changes requested):
Remove CORS allow credentials setting?

tituomin (dismissed):

tituomin (approved):

PR #161, opened 2021-03-09
Remove access_token hiding hack from helusername backend

vikoivun:
Remove a hack to work around a bug in python-jose and upgrade to a fixed version of python-jose.

Keycloak 9.x issued ID-tokens without at_hash claim during authorization code flow. This caused the buggy library to reject the token. See https://github.com/mpdavis/python-jose/issues/75

Newer versions of Keycloak started to include the claim even during authorization code flow, which masked the bug.

### DIFF
--- a/auth_backends/helsinki_username.py
+++ b/auth_backends/helsinki_username.py
@@ -9,6 +9,3 @@ class HelsinkiUsername(OpenIdConnectAuth):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.OIDC_ENDPOINT = self.setting('OIDC_ENDPOINT')
-
-    def validate_and_return_id_token(self, id_token, access_token):
-        super().validate_and_return_id_token(id_token, None)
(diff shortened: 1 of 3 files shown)


mikkokeskinen (changes requested):
Could the python-jose version be set in the requirements.in and running pip-compile?

Like so:
`python-jose>=3.2.0`

vikoivun:
> Could the python-jose version be set in the requirements.in and running pip-compile?
> 
> Like so:
> `python-jose>=3.2.0`

Well, it is a transitive dependency. I'm hoping social-core will get back into business and update their dependencies. At least the project recently got a new maintainer.

mikkokeskinen:
Yes I know. But if it's our need that the version should be newer, we should maybe set it in our requirements?

mikkokeskinen:
Btw. python-jose 3.2.0 brings other updates that are not now in the requirements.txt.

vikoivun:
> Btw. python-jose 3.2.0 brings other updates that are not now in the requirements.txt.

Fair enough. I'll push the changes to requirements.in

PR #164, opened 2021-03-25
Update ADFS signing certificate

vikoivun:
ADFS rolls its certificates every now and then. Tunnistamo does not
support dynamically retrieving them, so we need to update them
in the repo.


vikoivun:
Note: this change is already in production (ad-hoc patched) and the branch has been tested in sos-test.

vikoivun:
I'll look at the failing tests too.

tituomin:
> I'll look at the failing tests too.

Any reason not to assume the tests just fail for the obvious reason?

tituomin (approved):

PR #167, opened 2021-04-27
HP-630 Add multiple environment support for Yle

tituomin:
The valid environments are hardcoded to reduce
the scope of possible misconfiguration errors.


yletunnus/tests/test_backend.py
```diff
@@ -91,18 +95,46 @@ def authorize_body(self, request, url, headers):
...
+
+    def test_login_invalid_environment_raises_error(self):
+        for environment in INVALID_ENVIRONMENTS:
+            with self.subTest():
+                self.strategy.set_settings({
+                    'YLETUNNUS_ENVIRONMENT': environment,
+                    'SOCIAL_AUTH_USERNAME_IS_FULL_EMAIL': True
+                })
+                try:
+                    self.do_login()
```
mikkokeskinen, line 123: Could this be clearer as something like this?

```python
with pytest.raises(ImproperlyConfigured):
    self.do_login()
```

yletunnus/backends.py
```diff
@@ -10,8 +16,14 @@ class YleTunnusOAuth2(BaseOAuth2):
...
-    AUTHORIZATION_URL = 'https://auth.api.yle.fi/v1/authorize'
-    ACCESS_TOKEN_URL = 'https://auth.api.yle.fi/v1/token'
+
+    def _get_url(self, version='v1', *url_path):
+        environment = self.setting('YLETUNNUS_ENVIRONMENT')
+        if environment not in BASE_URLS.keys():
+            raise ImproperlyConfigured(
+                'Unknown YLE Tunnus environment. '
+                'Please set the settings variable SOCIAL_AUTH_YLETUNNUS_ENVIRONMENT to an allowed value.')
+        return '/'.join((BASE_URLS[environment], version) + url_path)
```
mikkokeskinen, line 26: Is there a good reason this just joins all the positional arguments to the url? In my opinion it would be clearer to drop the asterisk and check if the value of `url_path` is an iterable or not.

tunnistamo/settings.py
```diff
@@ -77,6 +77,9 @@
     SOCIAL_AUTH_YLETUNNUS_APP_KEY=(str, ""),
     SOCIAL_AUTH_YLETUNNUS_SECRET=(str, ""),
 
+    # Value must be either "test" or "production"
+    SOCIAL_AUTH_YLETUNNUS_ENVIRONMENT=(str, ""),
+
```
vikoivun, line 82: This looks to be the correct abstraction level.

vikoivun: Although, maybe default to production?

vikoivun (approved):

yletunnus/backends.py
```diff
@@ -24,7 +36,10 @@ def auth_params(self, state=None):
 
     def access_token_url(self):
         app_id, app_key = self.setting('APP_ID'), self.setting('APP_KEY')
-        return self.ACCESS_TOKEN_URL + '?app_id=%s&app_key=%s' % (app_id, app_key)
+        return self._get_url('token?app_id=%s&app_key=%s' % (app_id, app_key))
```
mikkokeskinen, line 39: I know this string formatting is as is. But could be changed to more modern?

```python
'token?app_id={}&app_key={}'.format(app_id, app_key)
# or
f'token?app_id={app_id}&app_key={app_key}'
```

tunnistamo/settings.py
tituomin: On the other hand, there will be more Tunnistamos connected to test than to prod. Hmm, which one would be more secure by default? Or insecure? Having production connected to test by accident or vice versa? This forces one to make the choice.

yletunnus/tests/test_backend.py
mikkokeskinen: Ah. Sorry this was a unittest class. Maybe should use [TestCase.assertRaises](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises)?

mikkokeskinen (commented):
These new fix commits could be squashed to the original commit?

yletunnus/tests/test_backend.py
akikoskinen: But aren't these tests still run with pytest? And pytest supports unittest classes, so can use `with pytest.raises`?

yletunnus/backends.py
```diff
@@ -10,8 +16,14 @@ class YleTunnusOAuth2(BaseOAuth2):
...
     DEFAULT_SCOPE = ['sub', 'email']
-    AUTHORIZATION_URL = 'https://auth.api.yle.fi/v1/authorize'
-    ACCESS_TOKEN_URL = 'https://auth.api.yle.fi/v1/token'
+
+    def _get_url(self, version='v1', *url_path):
+        environment = self.setting('YLETUNNUS_ENVIRONMENT')
+        if environment not in BASE_URLS.keys():
+            raise ImproperlyConfigured(
+                'Unknown YLE Tunnus environment. '
+                'Please set the settings variable SOCIAL_AUTH_YLETUNNUS_ENVIRONMENT to an allowed value.')
```
akikoskinen, line 25: Can the setting be checked earlier, and not during every time a URL is needed? Perhaps check the setting when this module gets imported?

yletunnus/tests/test_backend.py
mikkokeskinen: Of course one _can_, but I think it's neater to not mix and match unittest and pytest stuff. No need to import pytest then either.

yletunnus/tests/test_backend.py
```diff
@@ -91,18 +96,40 @@ def authorize_body(self, request, url, headers):
...
         self.complete_url = target_url
         HTTPretty.register_uri(HTTPretty.GET, start_url, body=self.authorize_body)
         return target_url
 
     def test_login(self):
-        self.strategy.set_settings({
-            'SOCIAL_AUTH_USERNAME_IS_FULL_EMAIL': True
-        })
-        self.do_login()
+        for environment in VALID_ENVIRONMENTS:
```
akikoskinen, line 108: Is it possible to use pytest's `@pytest.mark.parametrize` here?

yletunnus/tests/test_backend.py
```diff
@@ -91,18 +96,40 @@ def authorize_body(self, request, url, headers):
...
+            with self.subTest():
+                self.strategy.set_settings({
+                    'YLETUNNUS_ENVIRONMENT': environment,
+                    'SOCIAL_AUTH_USERNAME_IS_FULL_EMAIL': True
+                })
+                self.do_login()
+
+    def test_login_invalid_environment_raises_error(self):
+        for environment in INVALID_ENVIRONMENTS:
```
akikoskinen, line 117: Is it possible to use pytest's `@pytest.mark.parametrize` here?

mikkokeskinen: the subTest call is unittest way to do the same thing.

yletunnus/tests/test_backend.py
tituomin: > These new fix commits could be squashed to the original commit?

I will, in the end.

tituomin: 564e6e929356dd12285e052a1131813f8b8e41de fixes it with pytest, but I can do it with TestCase also.

yletunnus/backends.py
tituomin: Fixed in cd2b5f1a47945e6cf1a7b406532fbd4872a35c86 -- and at the same time it fixes a bug (the tests were incorrect too).

yletunnus/backends.py
tituomin: Fixed in 9df56232a4377d4d370197884f79abb013498df6

yletunnus/backends.py
tituomin: Yeah, I was thinking that we don't want to bother users who don't need this backend, but it seems that the module is loaded on-demand, so good idea.

yletunnus/tests/test_backend.py
tituomin: Yeah, I'm not 100% how mixing with pytest would work here, and maybe it's cleaner to use one style in one file/module?

yletunnus/backends.py
tituomin: Very difficult to test the checking in that case, though.

yletunnus/tests/test_backend.py
tituomin: > The following pytest features do not work, and probably never > will due to different design philosophies:
>
> - Fixtures (except for autouse fixtures, see below);
> - Parametrization;
> - Custom hooks;

https://docs.pytest.org/en/stable/unittest.html

yletunnus/backends.py
tituomin: @akikoskinen I managed to make the settings checking happen on module import, but with the Tunnistamo default AUTHENTICATION_BACKENDS having yle tunnus there, Django will fail to load because the backend is loaded on startup.

@mikkokeskinen @akikoskinen a more idiomatic way might be to use https://docs.djangoproject.com/en/3.2/topics/checks/ for this. What do you think?

mikkokeskinen: Unconfigured social backends haven't prevented starting the server thus far. I wouldn't change that. I think that we should set a default value for the YLE environment and check validity like it is now.

akikoskinen: If the check must happen only in the case when this backend is really _used_, then the check probably has to be within some method as it is now.

tunnistamo/settings.py
tituomin: @mikkokeskinen Which default? I'm thinking production, since our no1 aim is to protect the production instance and will never want to enable test there by accident.

mikkokeskinen: I would set it to "production" too.

yletunnus/backends.py
tituomin: I'm still thinking we should leverage the _System check framework_ for cases where we want to do some consistency checks wrt. our configuration values (for example: YLE tunnus is enabled as a backend, has id & secret configured but no environment). But maybe that's another ticket. Let's keep this as it is.

tunnistamo/settings.py
tituomin: Fixed in 48566927e92106c400af922b8361f0cf4a61d65b

yletunnus/tests/test_backend.py
tituomin: Fixed in 4bd071ad16607afa64bf7b14a09324b33a49b167

mikkokeskinen (approved):
Looking good! 👍

akikoskinen (commented):
Code looks fine for me, but I can't say whether this actually works, as I don't have a working development environment for this project.

tituomin:
> Code looks fine for me, but I can't say whether this actually works, as I don't have a working development environment for this project.

Tested locally, works.

PR #170, opened 2021-05-26
Fix get_loa not finding the correct loa value

mikkokeskinen:
The id_token in extra_data is encoded token, so we can't get the claims easily. Changed the get_loa to use the already decoded id_token.

The id_token value in the backend is saved only temporary when the OIDC backend validates the incoming id_token. It's enough because we only need the value in the pipeline. The reason I used extra_data was that the loa value could be read afterwards if we had a socialauth entry. Maybe we don't need that feature.

Also changed the test backend to actually generate an encoded id_token. Previously I cut corners when creating the test backend which ended up giving false postive test results.


akikoskinen (commented):
I'm quite clueless with this change. Seems legit as is and it could work. But that was my opinion already before this change. My problem is that I don't know the social auth library. I guess the concern I have is that quite a lot of the social auth functionality seems to be mocked in tests, so how much those tests ensure that things really work in reality? But I also don't know if there is another way.

Has this been deployed somewhere where it could be tested in real action?

akikoskinen (approved):
I'm confident that this works now :laughing:

PR #171, opened 2021-05-28
On social auth error redirect user to the OIDC client

mikkokeskinen:
After initiating an OIDC authorization, the user will select a login method and go to the social auth backend. If there is an error with the backend authentication, the user is redirected back to Tunnistamo and social auth raises an exception. The InterruptedSocialAuthMiddleware will catch the exception and always redirects the user back to the login method selector.

This patch changes the functionality so that the user is not redirected to the login method selector, but will be redirected back to the OIDC client they used. The original redirect_uri is appended with "error" and "error_description" parameters.


tunnistamo/middleware.py
```diff
@@ -1,24 +1,43 @@
...
 logger = logging.getLogger(__name__)
 
 
+def add_params_to_url(url, params):
+    """Add GET parameters to an existing url"""
+    url_parts = urlsplit(url)
+    new_query_parts = dict(parse_qsl(url_parts.query))
+    new_query_parts.update(params)
+    new_url_parts = list(url_parts)
+    new_url_parts[3] = urlencode(new_query_parts)
```
tituomin, line 25: Use named tuple syntax instead of hardcoded integer?

mikkokeskinen: The `url_parts` is a `SplitResult` which is immutable. The url_parts is converted to a list on the previous line.

I guess an alternative could be to create a new `SplitResult`. Something like this:

```python
return SplitResult(
    scheme=url_parts.scheme, 
    netloc=url_parts.netloc, 
    path=url_parts.path, 
    query=urlencode(new_query_parts),
    fragment=url_parts.fragment
).geturl()
```

akikoskinen: If `SplitResult` is a named tuple, as the documentation suggests, then it should have a [_replace](https://docs.python.org/3/library/collections.html#collections.somenamedtuple._replace) method.

Also, if you still have the `SplitResult` at hand, it should have a `geturl()` method, which returns the combined URL as string.

akikoskinen:
Some tests seem to fail in Travis. Temporary hiccup or something serious?

tunnistamo/middleware.py
mikkokeskinen: Changed the `add_params_to_url` function.

tunnistamo/tests/test_auth_error_redirect.py
```diff
@@ -0,0 +1,177 @@
...
+from users.models import LoginMethod
+
+
+def _build_full_url(extra, path):
+    return '{scheme}{host}{port}{path}'.format(
+        scheme=extra.get('wsgi.url_scheme') + '://' if extra.get('wsgi.url_scheme') else '',
+        host=extra.get('SERVER_NAME', ''),
+        port=':' + extra.get('SERVER_PORT') if extra.get('SERVER_PORT') else '',
+        path=path
+    )
```
akikoskinen, line 21: Could `urllib` be used here, instead of a hand-crafted solution?

tunnistamo/tests/test_auth_error_redirect.py
```diff
@@ -0,0 +1,177 @@
...
+        order=1,
+    )
+
+    redirect_uris = ['https://example.com/callback']
+    oidc_client = oidcclient_factory(redirect_uris=redirect_uris)
+
+    # This will redirect the user to the Tunnistamo login view and from there to the
+    # "Test login method". We cancel the last redirect in the
+    # CancelSocialAuthBackendRedirectClient. The cancelled redirect is required to
+    # have the "next" parameter in the django_clients session.
```
akikoskinen, line 46: While this comment is helpful, I think it's in a wrong location.

The first sentence (This will redirect the user to the Tunnistamo login view and from there to the
"Test login method".) feels like it should be this method's comment, so right below the `def _start_oidc_authorize...`.

The rest feels like it should be within the `CancelSocialAuthBackendRedirectClient` class, perhaps replacing the current comment that is there.

tunnistamo/tests/test_auth_error_redirect.py
```diff
@@ -0,0 +1,177 @@
...
+
+@pytest.mark.django_db
+@pytest.mark.parametrize('original_error,expected_error,expected_error_description', [
+    ('', 'interaction_required', _('Authentication failed')),
+    ('unknown error code', 'interaction_required', _('Authentication failed')),
+    ('interaction_required', 'interaction_required', _('Authentication failed')),
+    ('access_denied', 'access_denied', _('Authentication cancelled or failed')),
+])
+def test_redirect_to_client_after_social_auth_error(
+    django_user_model,
```
akikoskinen, line 87: `django_user_model` is not used here. Can it be removed?

tunnistamo/tests/test_auth_error_redirect.py
```diff
@@ -0,0 +1,177 @@
...
+    }))
+    assert response.status_code == 302
+    assert response.url == expected_redirect_url
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize('on_error_redirect', (True, False))
+def test_on_auth_error_redirect_to_client_setting(
+    settings,
+    django_user_model,
```
akikoskinen, line 120: `django_user_model` is not used here. Can it be removed?

tunnistamo/tests/test_auth_error_redirect.py
```diff
@@ -0,0 +1,177 @@
...
+        # the login view
+        assert response.url.startswith('/login/')
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize('on_error_redirect', (True, False))
+def test_should_not_redirect_to_oidc_client_if_the_next_parameter_is_not_to_an_oidc_endpoint(
+    settings,
+    django_user_model,
+    oidcclient_factory,
```
akikoskinen, line 158: `django_user_model` and `oidcclient_factory` are not used here. Can they be removed?

akikoskinen (commented):
A few small comments.

mikkokeskinen:
There were errors because I had local settings which made the tests pass locally but not on the server. I had to mock some endpoints with httpretty to get the social auth backend to generate the redirect url properly.

Also I decided to make the redirect cancelling more simple by just checking the domain name.

akikoskinen (approved):
Merge the two commits into one.

tituomin (approved):

akikoskinen (approved):

PR #172, opened 2021-06-01
Upgrade some requirements

mikkokeskinen:
Updated in this PR: django, Pillow, djangorestframework, lxml, and cryptography.

### DIFF
--- a/users/views.py
+++ b/users/views.py
@@ -89,7 +89,7 @@ class LoginView(TemplateView):
                     continue
                 if re.match(getattr(settings, 'SOCIAL_AUTH_SUOMIFI_CALLBACK_MATCH'), next_url) is None:
                     continue
-                m.login_url += '&amp;idp=' + m.provider_id
+                m.login_url += '&idp=' + m.provider_id
 
             methods.append(m)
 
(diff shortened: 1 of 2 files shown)


users/views.py
```diff
@@ -89,7 +89,7 @@ def get(self, request, *args, **kwargs):  # noqa  (too complex)
                     continue
                 if re.match(getattr(settings, 'SOCIAL_AUTH_SUOMIFI_CALLBACK_MATCH'), next_url) is None:
                     continue
-                m.login_url += '&amp;idp=' + m.provider_id
+                m.login_url += '&idp=' + m.provider_id
```
tituomin, line 92: What was this?

tituomin (approved):

users/views.py
mikkokeskinen: Some kind of strange copy-paste error I would assume. But it worked because that was equivalent to "&amp&idp=". In a newer Django the semicolon is no longer considered a query parameter separator, so the resulting parameter name was "amp;idp". Which obviously didn't work anymore.

tituomin (changes requested):
Since this is still open, upgrade to Django 2.2.24 while we're at it.

mikkokeskinen:
> Since this is still open, upgrade to Django 2.2.24 while we're at it.

Done.

tituomin (approved):

PR #176, opened 2021-08-19
Check Tunnistamo Session validity in OIDC views

mikkokeskinen:
Tunnistamo Session has a new field `ended_at` which will be set when the user logs out. The session will be considered invalid if the `ended_at` has a non-None value. The `ended_at` is set in a new user_logged_out-signal handler which gets triggered in the end session view (default log out uses this too). Backchannel log out ends the session too.

The session validity is checked in the following OIDC provider views:

- authorize view
- token view
- userinfo view
- token introspection view

Additionally the validity is checked in the API Tokens view.

Refs HP-592


auth_backends/backchannel_logout.py
```diff
@@ -101,5 +101,13 @@ def backchannel_logout(self, *args, **kwargs):
...
                 continue
 
-            logger.info(f'Deleted a session for user {session_user_id}')
+            if 'tunnistamo_session_id' in session_data:
+                from users.models import TunnistamoSession
+                try:
+                    tunnistamo_session = TunnistamoSession.objects.get(pk=session_data['tunnistamo_session_id'])
+                    tunnistamo_session.end()
+                except TunnistamoSession.DoesNotExist:
+                    pass
```
akikoskinen, line 110: What if the `django.contrib.auth.user_logged_out` signal would be sent here (somehow)? Would that be appropriate? I checked if `django.contrib.auth.logout` (which sends that signal) could be called, but it seems to have been made for cases where the logged in user **themself** makes a request to logout, as the `logout` function just receives a `request` as an argument and then pulls out a `user` and `session` from it. But wouldn't it be neat if the whole "logout machinery" would run on logout events, no matter where they originated from?

akikoskinen: I checked through the rest of the code and it's looking fine. So I suppose this is the only discussion topic that I can come up with (today :smirk:).

mikkokeskinen: Like you wrote, the signal handlers expect the request to be one the user themselves made. In this backchannel log out-case we would need to craft a fake request and somehow attach the appropriate Django session to it. I guess it could be possible, but I fear it would be too brittle.

akikoskinen: You are right. It probably can't be used in this case.

akikoskinen (approved):

tituomin (approved):
Prepend "HP-592" to commit messages. Otherwise good!

akikoskinen:
The `.travis.yml` file can probably be removed now? And if that doesn't remove the `travis-ci` check, then someone with enough privileges needs to get rid of it.

akikoskinen (approved):
So green now :+1:

PR #177, opened 2021-08-24
Add "sid" (Session ID) claim to id_token and API tokens

mikkokeskinen:
The value is the ID of the current Tunnistamo Session. The "sid" claim will be used in the RPs to match sessions and logout tokens.

Refs HP-811

### DIFF
--- a/oidc_apis/utils.py
+++ b/oidc_apis/utils.py
@@ -70,6 +70,9 @@ def additional_tunnistamo_id_token_claims(dic, user, token, request, **kwargs):
     if not tunnistamo_session:
         return dic
 
+    # Add Tunnistamo Session id as the "sid" (Session ID) claim
+    dic['sid'] = str(tunnistamo_session.id)
+
     # Set the social auth backend name as the "amr" (Authentication Methods Reference)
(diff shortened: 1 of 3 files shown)


oidc_apis/utils.py
```diff
@@ -70,6 +70,9 @@ def additional_tunnistamo_id_token_claims(dic, user, token, request, **kwargs):
     if not tunnistamo_session:
         return dic
 
+    # Add Tunnistamo Session id as the "sid" (Session ID) claim
+    dic['sid'] = str(tunnistamo_session.id)
```
akikoskinen, line 74: I don't see the comment bringing much value, as it pretty much repeats what the code says quite clearly already.

tituomin (approved):

oidc_apis/utils.py
mikkokeskinen: It's keeping up with the other comments. Also I think it's good to say what the "sid" means.

akikoskinen: Hmm, that's true. All the hard-to-understand claim names seem to have been documented there. Since that's the chosen path, then it's good to keep it that way.

akikoskinen (approved):

PR #178, opened 2021-08-31
OIDC Provider backchannel logout support

mikkokeskinen:
The backchannel logout is sent to APIs when the users Tunnistamo Session ends if the API has a backchannel logout url defined.

The logout is only sent to the APIs for which the user could have fetched an API token. i.e. The APIs which scopes are mentioned in the token.

Had to pass request-parameter deeper than I would have liked because oidc_providers `get_issuer()` needs to know what host name to use when generating the issuer value. Using request is the only way as we don't have `SITE_URL`-setting set.


auth_backends/backchannel_logout.py
```diff
@@ -105,7 +105,7 @@ def backchannel_logout(self, *args, **kwargs):
                 from users.models import TunnistamoSession
                 try:
                     tunnistamo_session = TunnistamoSession.objects.get(pk=session_data['tunnistamo_session_id'])
-                    tunnistamo_session.end()
+                    tunnistamo_session.end(send_logout_to_apis=True, request=self.request)
```
akikoskinen, line 108: Looks like there are no tests that check that when the back-channel logout endpoint of Tunnistamo is called, does Tunnistamo also call the other services' back-channel logout endpoints at that point.

tituomin (commented):
Prepend "HP-474" to commit messages?

tunnistamo/tests/test_oidc_provider_backchannel_logout.py
```diff
@@ -0,0 +1,203 @@
...
+
+    api_scope.api.backchannel_logout_url = 'invalid url'
+    api_scope.api.save()
+
+    send_backchannel_logout_to_apis_in_token_scope(token, request=RequestFactory().get('/'))
+
+
+@pytest.mark.django_db
+@httprettified
+def test_send_backchannel_logout_to_apis_one_api_in_scope(user):
```
akikoskinen, line 122: To me it seems this test doesn't cover anything more than the next test (`test_send_backchannel_logout_to_apis_two_apis_in_scope`), so perhaps this test could be removed?

akikoskinen (commented):
A few minor comments.

oidc_apis/backchannel_logout.py
```diff
@@ -0,0 +1,44 @@
...
+        'logout_token': create_logout_token(api.oidc_client, iss, sub, sid),
+    })
+    response.raise_for_status()
+
+
+def send_backchannel_logout_to_apis_in_token_scope(oidc_token, request, sid=None):
+    known_api_scopes = ApiScope.objects.by_identifiers(oidc_token.scope)
+    token_api_scopes = known_api_scopes.allowed_for_client(oidc_token.client)
+
+    for api_scope in token_api_scopes:
```
tituomin, line 29: There might be network timeouts or such things which might make this loop operation get stuck for a long while (visible to the user?). Also, in case of errors, do we ever want to retry? Would a job queue style more async implementation be more quick/robust?

mikkokeskinen: Yeah, maybe should add a short timeout value. Asynchronous requests would definitely be the correct choice. But we don't have Celery or anything like that in use in Tunnistamo.

auth_backends/backchannel_logout.py
mikkokeskinen: True. I'll need to add a test for that.

tunnistamo/tests/test_oidc_provider_backchannel_logout.py
mikkokeskinen: Maybe it's a philosophic question, but if we only test with two APIs, do we know that it works with only one API? 😄

users/models.py
```diff
@@ -260,10 +260,24 @@ def get_content_object_by_model(self, model_or_instance):
 
         return session_element.content_object
 
-    def end(self):
+    def end(self, send_logout_to_apis=False, request=None):
```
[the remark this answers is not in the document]
mikkokeskinen: I'd like to keep the default behavior not to call external services. As it's not really common in Django models.

tunnistamo/tests/test_oidc_provider_backchannel_logout.py
akikoskinen: It could very well be a valid question. If you think it's beneficial to have both of these tests, then it's ok to keep them.

users/models.py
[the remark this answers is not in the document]
akikoskinen: > ...call external services. As it's not really common in Django models

This text makes me ponder that is calling external services from a Django model the correct thing to do in **any** situation? Should this "ending a `TunnistamoSession`" functionality be extracted out of the model somehow?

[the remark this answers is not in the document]
mikkokeskinen: That's one thing I meant when I said I'd like to discuss this feature. But yeah, maybe. But it would get always called after the Tunnistamo Session is ended. Now the session is ended in two places, so it would get called in two different places. Of course if there would be more places in the future, one needs to remember to call the "API log out functionality" in those too.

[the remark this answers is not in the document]
akikoskinen: I propose there should be a "session ending" function somewhere, that does more or less all what is now in `TunnistamoSession.end`. I'm just having a hard time trying to come up with the correct location for it...

oidc_apis/backchannel_logout.py
tituomin: A short timeout value ok for now.

users/models.py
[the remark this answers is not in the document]
akikoskinen: Concluded that we'll let this be as it is, as no "best way" came about.

tituomin (approved):

tunnistamo/tests/test_oidc_provider_backchannel_logout.py
```diff
@@ -0,0 +1,294 @@
...
+
+
+def fix_httpretty_latest_requests_list(latest_requests):
+    """Deduplicate POST requests in HTTPretty's latest requests list
+
+    HTTPretty has introduced a bug where POST requests are doubled in the request
+    history. See Issue in their repository: https://github.com/gabrielfalcao/HTTPretty/issues/425"""
+    from httpretty import version as httpretty_version
+
+    if httpretty_version < '1.1' or httpretty_version > '1.1.4':
```
akikoskinen, line 27: 1.1.4 is currently the latest release of httpretty. It isn't certain that this issue will be fixed in the next release. Should here be some code that whoever updates httpretty dependency next, has to check what's the status of this issue? Something like `if httpretty_version > '1.1.4': assert False` (with an explanatory description what to do).

mikkokeskinen: Good point. I'll add it.

akikoskinen (approved):

PR #180, opened 2021-09-08
Redirect ADFS users to ADFS logout when logging out

mikkokeskinen:
If the user used an ADFS backend as their social login, redirect them to the ADFS logout url when they are logging out of Tunnistamo. `LOGOUT_URL` is only set in the Helsinki ADFS backend for now.

Refs HP-918

### DIFF
--- a/auth_backends/adfs/helsinki.py
+++ b/auth_backends/adfs/helsinki.py
@@ -8,6 +8,7 @@ class HelsinkiADFS(BaseADFS):
     name = 'helsinki_adfs'
     AUTHORIZATION_URL = 'https://fs.hel.fi/adfs/oauth2/authorize'
     ACCESS_TOKEN_URL = 'https://fs.hel.fi/adfs/oauth2/token'
+    LOGOUT_URL = 'https://fs.hel.fi/adfs/oauth2/logout'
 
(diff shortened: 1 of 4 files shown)


akikoskinen (commented):
Generally looks ok. But I'm not completely sure what exactly in technical terms this tries to achieve. For example, under what circumstances and at which point exactly in the logout flow this redirect should happen? Is there some standard that this flow is following?

Is setting the `next_page` attribute the correct thing to do? Or should there for example be a `return redirect(...)` line, as is used earlier in the `TunnistamoOidcEndSessionView.dispatch` method?

Because of the above I can't really judge if this is the correct implementation.

mikkokeskinen:
Yeah, that method is a bit hairy as there are so many different scenarios.

The `return redirect` is a bit different. There we don't want to log out just yet, just visit the OP end session and come back here to continue logging out.

In the ADFS case, we want to first log out from Tunnistamo (which is done in the `super().dispatch()`) and then end up in the ADFS log out.

akikoskinen:
All right, that explains it more and the implementation is now easier to understand. Would be good to add that information into the commit message.

tituomin (approved):

PR #183, opened 2021-09-16
Update some of the dependencies

mikkokeskinen:
Packages mentioned by snyk et al. Also couple that depend on those packages. (requests, python-jose)


akikoskinen:
Otherwise looks goody.

akikoskinen (approved):
Looks good.

PR #191, opened 2021-11-17
Improve HelsinkiTunnistus tests

akikoskinen:
Add a new test that ensures that the `original_client_id` parameter is passed on to the Helsinki tunnistus service. As preparation some test helpers were moved to a place where they can be used from many tests. Before this new test the [HelsinkiTunnistus.auth_extra_arguments](https://github.com/City-of-Helsinki/tunnistamo/blob/d676433f63866941b8786f1f06180b7aa4087a8a/auth_backends/helsinki_tunnistus_suomifi.py#L22-L36) method wasn't covered by any tests. It was possible to remove the method and no test would fail.

### DIFF
--- a/tunnistamo/tests/conftest.py
+++ b/tunnistamo/tests/conftest.py
@@ -1,24 +1,19 @@
 import json
-import time
 from urllib.parse import parse_qs, urlparse
 
 import jwt
 import pytest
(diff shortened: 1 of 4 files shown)


mikkokeskinen (approved):
Looks good.

PR #192, opened 2021-11-18
Obey user's desired UI language

akikoskinen:
Take into use [Django's regular locale middleware](https://docs.djangoproject.com/en/3.2/ref/middleware/#module-django.middleware.locale) that enables showing UIs to user's in their desired language. Expand that middleware to also consider a `ui_locales` query parameter in `GET` requests. Pass the user's UI language onward to Helsinki tunnistus authorization service, again as a `ui_locales` query parameter.

**NB**: Tunnistamo's UI isn't fully localized, so you could see an unexpected language there.

**NB2**: This is now version two of this feature. The first version is still available in the [HP-904-obey-users-language-old branch](https://github.com/City-of-Helsinki/tunnistamo/tree/HP-904-obey-users-language-old) so you can compare. The differences to the first version are shortly:

- The language cookie isn't used any more. The `ui_locales` value is stored into Django session and used from there.
  - As a consequence the `ui_locales` value is now passed forward to external services as is.
- The `ui_locales` can be a space separated list of language codes in priority order. That is now handled correctly.
- Remove some view specific `ui_locales` handling as that is now covered by the common code.

### DIFF
--- a/tunnistamo/middleware.py
+++ b/tunnistamo/middleware.py
@@ -31,11 +31,19 @@ class LocaleMiddleware(DjangoLocaleMiddleware):
     def process_request(self, request):
         ui_locales = request.GET.get('ui_locales')
 
-        try:
-            language = translation.get_supported_language_variant(ui_locales)
+        language = None
+        if ui_locales is not None:
+            for candidate in ui_locales.split():
+                try:
+                    language = translation.get_supported_language_variant(candidate)
+                    break
+                except LookupError:
+                    pass
+
+        if language:
             translation.activate(language)
             request.LANGUAGE_CODE = translation.get_language()
-        except LookupError:
+        else:
             super().process_request(request)
 
 
--- a/tunnistamo/middleware.py
+++ b/tunnistamo/middleware.py
@@ -5,8 +5,9 @@ from urllib.parse import parse_qsl, urlencode, urlsplit
 
 from django.conf import settings
 from django.http import HttpResponseForbidden, HttpResponseRedirect
+from django.middleware.locale import LocaleMiddleware as DjangoLocaleMiddleware
 from django.urls import reverse
-from django.utils import timezone
+from django.utils import timezone, translation
 from django.utils.http import quote
 from django.utils.translation import gettext_lazy as _
 from oidc_provider.lib.errors import BearerTokenError
@@ -24,6 +25,18 @@ def add_params_to_url(url, params):
     new_query_parts.update(params)
 
     return url_parts._replace(query=urlencode(new_query_parts)).geturl()
+
+
+class LocaleMiddleware(DjangoLocaleMiddleware):
+    def process_request(self, request):
+        ui_locales = request.GET.get('ui_locales')
+
+        try:
+            language = translation.get_supported_language_variant(ui_locales)
+            translation.activate(language)
+            request.LANGUAGE_CODE = translation.get_language()
+        except LookupError:
+            super().process_request(request)
 
 
 class InterruptedSocialAuthMiddleware(SocialAuthExceptionMiddleware):
--- a/tunnistamo/middleware.py
+++ b/tunnistamo/middleware.py
@@ -33,6 +33,8 @@ class LocaleMiddleware(DjangoLocaleMiddleware):
 
         language = None
         if ui_locales is not None:
+            request.session['ui_locales'] = ui_locales
+
             for candidate in ui_locales.split():
                 try:
                     language = translation.get_supported_language_variant(candidate)
--- a/tunnistamo/tests/test_locale_middleware.py
+++ b/tunnistamo/tests/test_locale_middleware.py
@@ -1,0 +1,12 @@
+import pytest
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize('lang', [None, 'fi', 'sv', 'en'])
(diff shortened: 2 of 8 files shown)


tituomin (approved):
I initially thought we should validate the incoming query parameter, but then realized Django validates it internally.

I noticed that sonarcloud complains about test coverage, maybe because it can't figure out the middleware is called?

tunnistamo/middleware.py
```diff
@@ -26,6 +28,33 @@ def add_params_to_url(url, params):
     return url_parts._replace(query=urlencode(new_query_parts)).geturl()
 
 
+class LocaleMiddleware(DjangoLocaleMiddleware):
+    def process_request(self, request):
+        try:
+            language = request.GET.get('ui_locales')
+            language = get_supported_language_variant(language)
+            translation.activate(language)
```
akikoskinen, line 36: This needs more work. The `ui_locales` can be a space separated list of language codes. The first usable should be chosen. I will fix this.

akikoskinen:
> I noticed that sonarcloud complains about test coverage, maybe because it can't figure out the middleware is called?

Yep, I don't know what its problem is. The Codecov.io report looks more correct.

tunnistamo/middleware.py
tituomin: Thank you!

akikoskinen: Fixed with the new version.

users/tests/conftest.py
```diff
@@ -29,6 +29,22 @@
...
+    changes the current active language."""
+    language_before_the_test = translation.get_language()
+
+    def reset_current_language():
+        if language_before_the_test:
+            translation.activate(language_before_the_test)
+        else:
+            translation.deactivate()
+
+    request.addfinalizer(reset_current_language)
```
mikkokeskinen, line 45: The `yield` style as recommended by the [pytest docs](https://docs.pytest.org/en/6.2.x/fixture.html#yield-fixtures-recommended) could be cleaner here?

users/tests/conftest.py
```diff
@@ -326,7 +326,9 @@ def get_json(self, url, *args, **kwargs):
     get_loa = HelsinkiTunnistus.get_loa
 
 
-def start_oidc_authorize(django_client, oidcclient_factory, backend_name=DummyFixedOidcBackend.name, state=None):
+def start_oidc_authorize(
```
mikkokeskinen, line 329: It might be nicer to have a separate `extra_data` argument and not just put all remaining arguments to `authorize_data`. Or explicitly just add a `ui_locales` parameter.

tunnistamo/middleware.py
```diff
@@ -26,6 +28,18 @@ def add_params_to_url(url, params):
     return url_parts._replace(query=urlencode(new_query_parts)).geturl()
 
 
+class LocaleMiddleware(DjangoLocaleMiddleware):
+    def process_request(self, request):
+        ui_locales = request.GET.get('ui_locales')
+
+        try:
+            language = get_supported_language_variant(ui_locales)
```
mikkokeskinen: The other functions from the `translation` module are called like `translation.function_name` but this is not. Maybe this could be called like that too?

tunnistamo/tests/test_locale_middleware.py
```diff
@@ -16,10 +17,17 @@ def test_language_is_determined_by_django_locale_middleware_if_ui_locales_provid
...
 
 
+NON_DEFAULT_LANGUAGE_CODE = list(filter(lambda lang: lang[0] != settings.LANGUAGE_CODE, settings.LANGUAGES))[0][0]
+
+
 @pytest.mark.django_db
-@pytest.mark.parametrize('ui_locales', ['fi', 'sv', 'en'])
-def test_language_is_determined_from_ui_locales_query_parameter(ui_locales, client, use_translations):
+@pytest.mark.parametrize('ui_locales,expected_lang', [
+    (f'{settings.LANGUAGE_CODE} {NON_DEFAULT_LANGUAGE_CODE}', settings.LANGUAGE_CODE),
```
mikkokeskinen, line 25: Could be a good idea to include a code with the region too? e.g. `fi-FI` or `en-US`

mikkokeskinen (changes requested):
Looks good. Couple of really minor things that could be changed in my opinion.

users/tests/conftest.py
akikoskinen: Yep. Changed.

users/tests/conftest.py
akikoskinen: True. Added separate `ui_locales` argument.

tunnistamo/middleware.py
akikoskinen: Indeed that was the case. How silly. Now all functions are used the same way.

tunnistamo/tests/test_locale_middleware.py
akikoskinen: Added. It's a non-existing specifier, but I guess Django doesn't care (and I guess it shouldn't be picky, because more specifiers can appear).

akikoskinen: I thought that it's not that important to test the `get_supported_language_variant` functionality that thoroughly here. Just trust that it does its job.

mikkokeskinen (approved):

PR #193, opened 2021-11-30
More translated views

akikoskinen:
Put more views under localization. The needed translations are marked with placeholders.

Also the docker image now contains the compiled translation files.

### DIFF
--- a/helsinki_theme/templates/helsinki_theme/base.html
+++ b/helsinki_theme/templates/helsinki_theme/base.html
@@ -1,4 +1,4 @@
-{% load static sass_tags %}
+{% load static sass_tags i18n %}
 <!DOCTYPE html>
 <html{% if request.LANGUAGE_CODE %} lang="{{ request.LANGUAGE_CODE }}"{% endif %}>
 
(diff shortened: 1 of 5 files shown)


helsinki_theme/templates/helsinki_theme/base.html
```diff
@@ -19,7 +19,11 @@
     <div class="col-sm-8 col-sm-offset-2 col-md-6 col-md-offset-3">
       <div class="content-card">
         <header>
-          <img src="{% static "hel-bootstrap-3/src/assets/helsinki-logo-black.svg" %}" class="logo" alt="Helsinki Helsingfors">
+          {% if request.LANGUAGE_CODE == "sv" %}
```
mikkokeskinen, line 22: Could maybe use the `get_current_language` template tag here?

mikkokeskinen (changes requested):

helsinki_theme/templates/helsinki_theme/base.html
akikoskinen: Changed. How does it look?

mikkokeskinen (approved):
👍

PR #194, opened 2021-12-01
Set static file serving dynamically from environment variables

akikoskinen:
The uWSGI configuration can use the already existing Django settings, as long as they are given as environment variables.

This uses uWSGI's [environment variable substitution feature](https://uwsgi-docs.readthedocs.io/en/latest/ParsingOrder.html#expanding-variables-placeholders).

### DIFF
--- a/.prod/uwsgi.ini
+++ b/.prod/uwsgi.ini
@@ -2,8 +2,8 @@
 http-socket = :8000
 chdir = /app
 module = tunnistamo.wsgi
-static-map = /media=/app/media
-static-map = /sso/static=/var/tunnistamo/static
+static-map = $(MEDIA_URL)=$(MEDIA_ROOT)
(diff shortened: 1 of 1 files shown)


vikoivun (dismissed):
Ok. Hopefully there aren't any problems with missing slashes before or after urls/paths.

akikoskinen:
> Ok. Hopefully there aren't any problems with missing slashes before or after urls/paths.

Hmm, yeah. This solution requires that `MEDIA_URL` and `STATIC_URL` begin with a slash. uWSGI seems to not be too picky with trailing slashes.

`MEDIA_ROOT` and `STATIC_ROOT` should be absolute paths according to Django anyway, so those should be good.

mikkokeskinen (approved):

PR #195, opened 2021-12-07
Consider ui_locales from session for determining current language

akikoskinen:
The `ui_locales` value was already stored in the session. But it was not used for determining the user's desired language during that session. This commit adds that.

The language determination logic is now thus:
- Use `ui_locales` query parameter
- Use `ui_locales` stored in session (added in this commit)
- Use the default Django locale middleware's functionality

The use case for this is the following. A user wants to authenticate. They enter the `/openid/authorize` endpoint, providing their desired language with a `ui_locales` query parameter. Tunnistamo responds with a 302 code and redirects the client to the `/login` page. This login page should be shown with the language the user desired. Before this PR this didn't happen.

### DIFF
--- a/tunnistamo/middleware.py
+++ b/tunnistamo/middleware.py
@@ -31,10 +31,14 @@ class LocaleMiddleware(DjangoLocaleMiddleware):
     def process_request(self, request):
         ui_locales = request.GET.get('ui_locales')
 
-        language = None
         if ui_locales is not None:
             request.session['ui_locales'] = ui_locales
+        else:
+            ui_locales = request.session.get('ui_locales')
 
+        language = None
+
+        if ui_locales:
             for candidate in ui_locales.split():
                 try:
                     language = translation.get_supported_language_variant(candidate)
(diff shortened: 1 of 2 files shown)


tunnistamo/middleware.py
```diff
@@ -31,10 +31,14 @@ class LocaleMiddleware(DjangoLocaleMiddleware):
     def process_request(self, request):
         ui_locales = request.GET.get('ui_locales')
```
mikkokeskinen, line 32: Could this be read more easily if it was:

```python
        ui_locales = request.GET.get('ui_locales')
        if ui_locales is None:
            ui_locales = request.session.get('ui_locales')
        else:
            request.session['ui_locales'] = ui_locales
```

I guess it's matter of taste and not too important.

mikkokeskinen (approved):
Looks good.

tunnistamo/middleware.py
akikoskinen: I don't know. I kind of prefer how it was, feels more natural order to me. But the difference (if there is any) is minuscule.

PR #196, opened 2021-12-09
Make Django session & csrf cookie settings configurable from env

vikoivun:


tunnistamo/settings.py
```diff
@@ -354,6 +359,13 @@
...
 
+CSRF_COOKIE_NAME = '{}-csrftoken'.format(env('COOKIE_PREFIX'))
+CSRF_COOKIE_PATH = env('PATH_PREFIX')
+CSRF_COOKIE_SECURE = env('COOKIE_SECURE')
+SESSION_COOKIE_NAME = '{}-sessionid'.format(env('COOKIE_PREFIX'))
+SESSION_COOKIE_PATH = env('PATH_PREFIX')
+SESSION_COOKIE_SECURE = env('COOKIE_SECURE')
+
 CSRF_COOKIE_NAME = 'sso-csrftoken'
 SESSION_COOKIE_NAME = 'sso-sessionid'
```
mikkokeskinen, line 370: Might want to remove these.

mikkokeskinen (changes requested):

tunnistamo/settings.py
vikoivun: Fixed

tunnistamo/settings.py
```diff
@@ -29,6 +29,11 @@
     MEDIA_URL=(str, '/media/'),
     NODE_MODULES_ROOT=(str, os.path.join(BASE_DIR, 'node_modules')),
 
+    COOKIE_SECURE=(bool, True),
+    COOKIE_PREFIX=(str, 'tunnistamo'),
+    # For scoping cookies to path
+    PATH_PREFIX=(str, '/'),
```
mikkokeskinen, line 35: Now that I read this again, I think this name might be too ambiguous. I mean, if you see that in the Azure library, you wouldn't know which path that affects. So maybe `COOKIE_PATH_PREFIX`?

akikoskinen: Or just `COOKIE_PATH` as the property's name in the cookie is `Path`?

vikoivun: I picked that up from HELERM and decided to use that for some consistency. But is indeed too generic. Will change.

mikkokeskinen: I guess Ville named it prefix as it's used in two settings and they in turn might set different value after the prefix. Which is improbable but possible. But in that case there should be two environment values for those too. :)

tunnistamo/settings.py
```diff
@@ -29,6 +29,11 @@
     MEDIA_URL=(str, '/media/'),
     NODE_MODULES_ROOT=(str, os.path.join(BASE_DIR, 'node_modules')),
 
+    COOKIE_SECURE=(bool, True),
+    COOKIE_PREFIX=(str, 'tunnistamo'),
```
akikoskinen, line 33: Should this be named `COOKIE_NAME_PREFIX`?

tunnistamo/settings.py
vikoivun: Just copied this from helerm without thinking. Technically it is prefix, as per http specification (or MDN at least). But it is not called a prefix

vikoivun: I'm not invested in the name, just wanted to have them under a single parameter as I've never seen them setup to separate values.

mikkokeskinen: `CSRF_AND_SESSION_COOKIE_PATH` then? 😄

vikoivun: That is rather long and thus typo prone. How about:
DJANGO_COOKIE_NAME_PREFIX : name prefix for all Django cookies
DJANGO_COOKIE_PATH: path for all Django cookies
DJANGO_COOKIE_SECURE: secure flag for all Django cookies
right now those cookies are {prefix}-sessionid and {prefix}-csrf

I'll also volunteer to push them into our documentation on dev.hel.fi, which also is in need of some care.

akikoskinen: Perhaps the `DJANGO_` prefix isn't needed as it's not used in any other name either.

vikoivun: Fair enough. I'll shorten them.

akikoskinen: Otherwise I liked the names just as proposed: COOKIE_NAME_PREFIX, COOKIE_PATH, COOKIE_SECURE. Those are my favorites :smile:

tunnistamo/settings.py
vikoivun: Done

tunnistamo/settings.py
vikoivun: Names are changed as per conversation.

akikoskinen:
One cookie related test seems to fail now.

mikkokeskinen:
Yes, the tests should use the cookie name setting and not rely on hard-coded value.

mikkokeskinen (approved):
👍

PR #198, opened 2021-12-09
Add ability to give command line arguments to uwsgi

akikoskinen:
**Edit**: #194 was chosen instead of this PR.

Extra command line arguments can be given with `$ARGS_UWSGI` environment variable. The name is chosen like this because uwsgi itself uses environment variables starting with `UWSGI_`.

This feature would make #194 obsolete as that, and much more, can be achieved with this more general solution.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -92,5 +92,5 @@ if [[ ! -z "$@" ]]; then
 elif [[ "$DEV_SERVER" = "1" ]]; then
     python ./manage.py runserver 0.0.0.0:8000
 else
-    uwsgi --ini .prod/uwsgi.ini
+    uwsgi --ini .prod/uwsgi.ini $ARGS_UWSGI
 fi


mikkokeskinen:
I wouldn't move the media and static settings to this. As it would be yet another thing to remember. (If I understood correctly)

Also, If there is no reason to keep the name as short as possible I would maybe make the name more descriptive. `EXTRA_COMMAND_LINE_ARGUMENTS_FOR_PRODUCTION_UWSGI` is maybe too long, but something like that.

akikoskinen:
> I wouldn't move the media and static settings to this. As it would be yet another thing to remember. (If I understood correctly)

Do you have a proposal how to do it? The #194 proposal is incomplete, as it just doesn't work for some valid Django setting values, namely values where the `*_URL` setting doesn't start with a slash.

mikkokeskinen:
Ah. I thought the #194 was working and this was for some other thing.

vikoivun:
At least the case of "no-preceding-url", which prepends script_name to the url is a kind of niche case. I don't know how you would provide non-hardcoded script_name to any program serving the static or media files. And if it is hardcoded you could just add it to STATIC_URL.

akikoskinen:
So. Do we choose to go with #194 after all and require that `*_URL` settings always start with a slash? (There could off course be some check somewhere too, that prints a warning or refuses to start the server if that's not the case)

akikoskinen:
#194 was chosen instead of this PR.

PR #206, opened 2022-01-17
Support sub-path serving in the Docker image

vikoivun:
Use uWSGI "mount"-mechanism to support serving tunnistamo from paths other than root.

### DIFF
--- a/.prod/uwsgi.ini
+++ b/.prod/uwsgi.ini
@@ -1,7 +1,8 @@
 [uwsgi]
 http-socket = :8000
 chdir = /app
-module = tunnistamo.wsgi
+mount = $(APP_URL_PATH)=tunnistamo/wsgi.py
+manage-script-name = true
 static-map = $(MEDIA_URL)=$(MEDIA_ROOT)
 static-map = $(STATIC_URL)=$(STATIC_ROOT)
 uid = appuser
--- a/Dockerfile
+++ b/Dockerfile
@@ -60,6 +60,7 @@ ENTRYPOINT ["./docker-entrypoint.sh"]
 # STore static files under /var to not conflict with development volume mount
 ENV STATIC_ROOT /var/tunnistamo/static
 ENV NODE_MODULES_ROOT /var/tunnistamo/node_modules
+ENV APP_URL_PATH /
 COPY --from=staticbuilder --chown=appuser:appuser /app/static /var/tunnistamo/static
 COPY --from=staticbuilder --chown=appuser:appuser /app/node_modules /var/tunnistamo/node_modules
 


Dockerfile
```diff
@@ -60,6 +60,7 @@ ENTRYPOINT ["./docker-entrypoint.sh"]
 # STore static files under /var to not conflict with development volume mount
 ENV STATIC_ROOT /var/tunnistamo/static
 ENV NODE_MODULES_ROOT /var/tunnistamo/node_modules
+ENV APP_URL /
```
akikoskinen, line 63: If I'm not mistaken this isn't really a URL, just a path. Something like `/path/to/my/app`. The variable could be named accordingly.

vikoivun: It is true that this can only be URL path as opposed to STATIC_URL and MEDIA_URL which can be full URLs. Would APP_URL_PATH be a good name?

akikoskinen: I suppose that works.

akikoskinen (changes requested):
Squash the commits and we are good to go.

vikoivun:
> Squash the commits and we are good to go.

I'll just that check build&deploy to 'shift succeeds.

vikoivun:
> > Squash the commits and we are good to go.
> 
> I'll just that check build&deploy to 'shift succeeds.

And it does. I've now squashed the commits.

akikoskinen (changes requested):
Commit message still mentions the old variable name.

.prod/uwsgi.ini
```diff
@@ -1,7 +1,8 @@
 [uwsgi]
 http-socket = :8000
 chdir = /app
-module = tunnistamo.wsgi
+mount = $(APP_URL_PATH)=tunnistamo/wsgi.py
+manage-script-name = true
 static-map = $(MEDIA_URL)=$(MEDIA_ROOT)
 static-map = $(STATIC_URL)=$(STATIC_ROOT)
```
akikoskinen, line 7: How these work together? If there's a `mount` now with a non-empty `$APP_URL_PATH`, are the `static-maps` checked with a path where the `$APP_URL_PATH` has been removed? Or are `static-maps` checked before the `mount`, with the original path?

akikoskinen, line 7: How these work together? If there's a `mount` now with a non-empty `$APP_URL_PATH`, are the `static-maps` checked with a path where the `$APP_URL_PATH` has been removed? Or are `static-maps` checked before the `mount`, with the original path?

vikoivun: First option. Ie.
APP_URL_PATH=/sso-test
STATIC_URL=/static
results in statics being available at https://{hostname}/sso-test/static
Hmm. I think I should still test that this works when APP_URL_PATH is at root (/). In principle that leads to double slashes in path.

akikoskinen: So does the `manage-script-name = true` then make Django have `$SCRIPT_NAME` set to `$APP_URL_PATH`, and when it thus generates static URLs, it basically builds them as `$SCRIPT_NAME/$STATIC_URL`, which would effectively be the same as `$APP_URL_PATH/$STATIC_URL`?

vikoivun: That is what happens indeed. Although I'm now confused as that should only happen when STATIC_URL is set to a relative path:
https://docs.djangoproject.com/en/4.0/ref/settings/#static-url
I'll do more testing.

vikoivun: All right. I've now tested and the behavior is as documented. I just was careless while testing.

vikoivun: Also, STATIC_URL and MEDIA_URL must contain the prefix for uWSGI to serve them correctly.  uWSGI does not have the interaction between static-maps and manage-script-name, as I previously erroneously claimed.

akikoskinen: So with the current version of uwsgi.ini in this PR, for uWSGI to serve the static files correctly, is it required to set:
APP_URL_PATH=/sso-test
STATIC_URL=/sso-test/static

Does that then make Django generate static URLs starting with /sso-test/sso-test/static (SCRIPT_NAME and STATIC_URL concatenated)? That would be incorrect.

akikoskinen (changes requested):
Can we do rebases instead of merge commits, please?

.prod/uwsgi.ini
vikoivun: No. The path will only contain "/sso-test" once. Django only concatenates SCRIPT_NAME and STATIC_URL when the latter is a relative URL. As per the documentation above and also per my testing.

vikoivun:
> Can we do rebases instead of merge commits, please?

Yes. The button on GitHub is a trap.

akikoskinen:
> > Can we do rebases instead of merge commits, please?
> 
> Yes. The button on GitHub is a trap.

Yep. The usage of those should be prevented :laughing:

.prod/uwsgi.ini
akikoskinen: Aha, I might have misunderstood what is meant with "relative path" in case of URLs. I thought that anything that doesn't mention the protocol and domain parts of a URL is considered relative. But it seems it means only the path part and whether that is relative or not (with the Unix interpretation of paths that start from the root directory `/` I suppose – although file system paths and URL paths are different beasts altogether).

akikoskinen (approved):

PR #207, opened 2022-01-24
Don't use random strings in test parametrization

akikoskinen:
The full test names generated by Pytest, which include the parameter values, change all the time if the parameters are random. That can e.g. mess up IDEs as they have hard time keeping track what are the tests' names.

### DIFF
--- a/tunnistamo/tests/test_oidc_provider_pkce.py
+++ b/tunnistamo/tests/test_oidc_provider_pkce.py
@@ -11,7 +11,7 @@ from oidc_provider.lib.utils.token import create_code
 @pytest.mark.django_db
 @pytest.mark.parametrize('code_verifier,should_succeed', (
     ('', False),
-    (get_random_string(), True),
+    pytest.param(get_random_string(), True, id='random_string'),
 ))
 def test_token_endpoint_requires_code_verifier(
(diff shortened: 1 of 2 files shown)


tunnistamo/tests/test_oidc_provider_pkce.py
```diff
@@ -11,7 +11,7 @@
 @pytest.mark.django_db
 @pytest.mark.parametrize('code_verifier,should_succeed', (
```
mikkokeskinen, line 12: I'd prefer to give them names instead of not using a random value. e.g.

```python
@pytest.mark.parametrize('code_verifier,should_succeed', (
    ('', False),
    (get_random_string(), True),
), ids=['empty_string', 'random_string'])
```

mikkokeskinen (changes requested):

tunnistamo/tests/test_oidc_provider_pkce.py
akikoskinen: Did that. Although with a slightly different syntax, only giving names for the random parameter tests and letting Pytest handle the rest.

mikkokeskinen (approved):

PR #208, opened 2022-01-25
Make end user visible texts for login methods translatable

akikoskinen:
Django-parler has this [documentation](https://django-parler.readthedocs.io/en/stable/advanced/migrating.html) for turning a previously non-translated field into a translated one. The problem with that document is, in reality [it](https://github.com/django-parler/django-parler/issues/154) [doesn't](https://github.com/django-parler/django-parler/issues/254) [work](https://github.com/django-parler/django-parler/issues/263) like that.

See commit messages for details how it was solved.


akikoskinen:
> It could be nice to group the translated and non-translated fields in the LoginMethodAdmin view like it's done in the `services.admin.ServiceAdmin`.

Good idea. Did that.

mikkokeskinen (approved):

PR #210, opened 2022-02-01
Don't use secure cookies for local dev environment

charn:
Local (docker based) development environment doesn't work by default if secure cookies are enforced.


akikoskinen (approved):

charn:
I don't have enough permissions to merge this. :sob:

nikomakela (approved):
I can approve that the fix is good!

akikoskinen:
No voi kökkö.

I probably would have the needed privileges. But there's a required action failing which prevents merging. Let's see if I can persuade this to pass somehow...

PR #212, opened 2022-02-09
Make STATIC_URL & MEDIA_URL work in container

vikoivun:
As it turns out uWSGI has interaction with mount-mechanism and static-map mechanism. With following configuration:

> mount = /sso-test=tunnistamo/wsgi.py
> static-map = /static=/var/tunnistamo/static

The file
`/var/tunnistamo/static/css/helsinki_theme.css`
will be available at URL
`https://hostname/sso-test/static/css/helsinki_theme.css`

So this means that uWSGI needs to be told to serve `/static` and Django needs to be told to generate URLs to `/sso-test/static/`. This PR attempts to that.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -92,5 +92,20 @@ if [[ ! -z "$@" ]]; then
 elif [[ "$DEV_SERVER" = "1" ]]; then
     python ./manage.py runserver 0.0.0.0:8000
 else
+    # We want to have a *_URL environment variables configure
+    # both Django static/media files URL generation
+    # and the corresponding file serving in uWSGI.
+    # uWSGI combines app mount path (in APP_URL_PATH) and static
+    # map path (in *_URL). For Django we combine them here.
+
+    # remove trailing slash for uWSGI URLs. uWSGI probably
+    # does this by itself, but it does not hurt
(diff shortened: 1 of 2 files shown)


akikoskinen:
The second commit should be squashed into the first one.

vikoivun:
> The second commit should be squashed into the first one.

Indeed. I'm still hoping to do that at a cursory local test.

docker-entrypoint.sh
```diff
@@ -92,5 +92,20 @@ if [[ ! -z "$@" ]]; then
...
+    # and the corresponding file serving in uWSGI.
+    # uWSGI combines app mount path (in APP_URL_PATH) and static
+    # map path (in *_URL). For Django we combine them here.
+
+    # remove trailing slash for uWSGI URLs. uWSGI probably
+    # does this by itself, but it does not hurt
+    export STATIC_URL_ROOTLESS=$(readlink -m $STATIC_URL)
+    export MEDIA_URL_ROOTLESS=$(readlink -m $MEDIA_URL)
+    # Use readlink to remove double initial slashes in case of
+    # APP_URL="/" and STATIC_URL="/static/". Also Django
```
akikoskinen, line 106: `s/APP_URL/APP_URL_PATH` in the comment.

akikoskinen (approved):

PR #222, opened 2022-03-01
Add azure ad for vantaa

mikkokeskinen:
Refs HP-1287


akikoskinen (commented):
Looks good to me, but hard to say anything about the correctness, as I'm not really sure what Vantaa is requesting :roll_eyes:

akikoskinen (approved):
I guess I can as well approve this, although I'm not sure if this pleases the Vantaa people. But perhaps it is to be found out later on.

PR #227, opened 2022-03-15
Update dependencies and Python version

akikoskinen:
To mitigate vulnerable code. Pillow update required Python 3.8, so had to update Python version too.

Closes #201
Closes #202
Closes #204
Closes #211
Closes #215
Closes #216
Closes #226


mikkokeskinen (approved):
flake8 suddenly marks one test as too complex, but I guess we can change it later.

akikoskinen:
> flake8 suddenly marks one test as too complex, but I guess we can change it later.

Yep. There's actually a [comment in that test](https://github.com/City-of-Helsinki/tunnistamo/blob/7295432566e8aa4a4b25a54a03e47ce26e95ef0c/users/tests/test_cors_middleware.py#L47) that supposedly previously made flake8 skip this test function for that check. Perhaps it's good to check if that function could be simplified – it does really look somewhat complex.

PR #229, opened 2022-03-18
Split one mammoth test into multiple smaller ones

akikoskinen:
The previous `test_cors_headers_got_with_whitelisted_uris_apiendpoints` test checked EVERYTHING related to the CORS headers. It was executed with various set of arguments and its behaviour was different according to those. It also contained needlessly complex logic that was very hard to read and understand.

Extracted the essence of that mammoth into smaller test functions. Now each test function checks one thing only. The meaning of each test function is now also documented in the function names.

### DIFF
--- a/tunnistamo/settings.py
+++ b/tunnistamo/settings.py
@@ -348,7 +348,7 @@ LOGGING = {
     }
 }
 
-CORS_ORIGIN_ALLOW_ALL = False
+CORS_ALLOW_ALL_ORIGINS = False
 CORS_URLS_REGEX = r'.*/(\.well-known/openid-configuration|v1|openid|api-tokens|jwt-token).*'
 
 
--- a/users/tests/conftest.py
+++ b/users/tests/conftest.py
@@ -98,7 +98,7 @@ def application_factory():
         kwargs.setdefault('name', get_random_string())
         kwargs.setdefault('client_id', get_random_string())
         kwargs.setdefault('user', None)
-        kwargs.setdefault('redirect_uris', None)
+        kwargs.setdefault('redirect_uris', '')
         kwargs.setdefault('client_type', Application.CLIENT_PUBLIC)
         kwargs.setdefault('authorization_grant_type', Application.GRANT_IMPLICIT)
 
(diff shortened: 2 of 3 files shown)


mikkokeskinen (commented):
I think the previous test also tested that the redirect_uris got saved to the AllowedOrigins  correctly. Do you plan on adding such tests again or...?

akikoskinen:
> I think the previous test also tested that the redirect_uris got saved to the AllowedOrigins correctly. Do you plan on adding such tests again or...?

No, I didn't plan to check the state of the database as that to me looked like testing an implementation detail. The AllowedOrigins is a way to implement the functionality that determines whether to add CORS headers to a request. The functionality has been tested, not the internal implementation details.

mikkokeskinen:
In my understanding, the feature is "Add Allow Origin (and couple other) headers for all of the domains that are listed in the clients redirect uris".

The other half of the feature is surely the parsing of all of the redirect uris correctly and adding/removing them from the whitelist.

akikoskinen:
> In my understanding, the feature is "Add Allow Origin (and couple other) headers for all of the domains that are listed in the clients redirect uris".

Also OAuth applications in addition to OIDC clients. And also post logout URIs in addition to redirect URIs. But yeah, that's the feature.

> The other half of the feature is surely the parsing of all of the redirect uris correctly and adding/removing them from the whitelist.

Yes, that's really important. And it is tested. But instead of examining the database state directly, it's tested by _using_ the whitelist for its intended purpose. I.e. making requests with various origins and checking if CORS headers are added correctly or not. Also clients (and OAuth applications) are added, modified and deleted to check that all those changes also affect the whitelist. So, testing _functionality_ instead of _implementation_.

mikkokeskinen:
Somehow I missed the URI_1 and URI_2 thingies. I guess I was looking for more complex redirect uris like in the old test. Maybe one path test is enough.

mikkokeskinen (dismissed):

akikoskinen:
Rebased to current `develop` branch.

mikkokeskinen (approved):

PR #240, opened 2022-06-01
Fetch users AD groups from the Microsoft Graph

mikkokeskinen:
In cases where the user had over 200 groups in Azure AD the provided claims didn't include "groups" claim at all, but included "Groups overage claim". In which case the helusers update_ad_groups-method used in
`users.pipeline.update_ad_groups` pipeline stage failed because it expects the claim to always being a list.

See https://docs.microsoft.com/en-us/azure/active-directory/develop/id-tokens#groups-overage-claim
for more information about the Group Overage Claim.

Example of Group Overage Claim:

```json
    "_claim_names": {
        "groups": "src1"
    },
    "_claim_sources": {
        "src1": {
            "endpoint": "https://graph.windows.net/00000000-0000-0000-0000-000000000000/users/00000000-0000-0000-0000-000000000000/getMemberObjects"
        }
    }
```

The endpoint mentioned in the `_claim_sources` doesn't seem to work with the access tokens available when logging in. Neither did the tokens acquired by using the `AzureADV2TenantOAuth2` backend. But the tokens acquired by it work with
Microsoft Graph.

It was then decided (with Ville K) to get the AD groups from the Microsoft Graph
in all cases even when the claims include the "groups" claim.

There are also a couple of tests for this functionality. They use the tests from
the social_core as a base.

Refs HP-1388

### DIFF
--- a/auth_backends/helsinki_azure_ad.py
+++ b/auth_backends/helsinki_azure_ad.py
@@ -1,9 +1,13 @@
+import logging
 from functools import lru_cache
+from json import JSONDecodeError
 
 from django.utils.translation import gettext, pgettext
 from requests import RequestException
 from social_core.backends.azuread_tenant import AzureADV2TenantOAuth2
 from social_core.exceptions import AuthFailed
+
+logger = logging.getLogger(__name__)
 
 
 class HelsinkiAzureADTenantOAuth2(AzureADV2TenantOAuth2):
@@ -14,6 +18,7 @@ class HelsinkiAzureADTenantOAuth2(AzureADV2TenantOAuth2):
     name_logged_in_to = pgettext('logged in to []', 'Azure AD')
     name_logout_from = pgettext('log out from []', 'Azure AD')
     name_goto = pgettext('go to []', 'Azure AD')
+    USER_DATA_URL = 'https://graph.microsoft.com/v1.0/me/memberOf?$select=id,displayName,securityEnabled'
 
     @property
     @lru_cache()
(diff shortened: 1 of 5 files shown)


auth_backends/helsinki_azure_ad.py
```diff
@@ -1,19 +1,24 @@
...
 from requests import RequestException
-from social_core.backends.azuread_tenant import AzureADTenantOAuth2
+from social_core.backends.azuread_tenant import AzureADV2TenantOAuth2
 from social_core.exceptions import AuthFailed
 
+logger = logging.getLogger(__name__)
 
-class HelsinkiAzureADTenantOAuth2(AzureADTenantOAuth2):
+
+class HelsinkiAzureADTenantOAuth2(AzureADV2TenantOAuth2):
```
akikoskinen, line 13: I suppose the base class change changes the version of the Microsoft OAuth2 API that is used or something like that. This change could perhaps be in its own commit, with an explanation why it's done.

It also seems that the new base class `AzureADV2TenantOAuth2` overrides a couple methods (`get_user_id` and `get_user_details`) from its base class and seems to change behavior somewhat. Is that OK?

akikoskinen (commented):
A few comments, nothing major.

akikoskinen:
Microsoft has done this "Groups overage claim" feature with the reason that the system wouldn't generate too large tokens to be put into HTTP headers. I guess we are not concerned about that and are putting all the groups into tokens, not minding any limits HTTP headers might have?

I know it's a difficult issue, as there is no standard upper limit to an HTTP header. IIRC HTTP standard doesn't have any size limitations for HTTP headers, but different server implementations can have their own (usually configurable) limits. Is the plan forward to just go and raise those configurable limits in any services that might hit a limit?

auth_backends/tests/test_helsinki_azure_ad.py
```diff
@@ -0,0 +1,130 @@
+import json
+
+from auth_backends.tests.azure_ad_base import AzureADV2TenantOAuth2Test, _generate_access_token_body
+
+
+class TestNoErrorWhenADGroupsNotInTokenOrGraph(AzureADV2TenantOAuth2Test):
```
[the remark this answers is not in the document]
mikkokeskinen: It's because the class attribute access_token_body is changed between tests. Maybe could hack it somehow, but I don't immediately see the value.

auth_backends/helsinki_azure_ad.py
mikkokeskinen: Separated the change to it's own commit. Yeah the changes are ok.

auth_backends/tests/azure_ad_base.py
```diff
@@ -0,0 +1,186 @@
...
+                'users.pipeline.get_username',
+                # Not compatible with the tests: 'users.pipeline.require_email',
+                'users.pipeline.associate_by_email',
+                'social_core.pipeline.user.create_user',
+                'users.pipeline.check_existing_social_associations',
+                'social_core.pipeline.social_auth.associate_user',
+                'social_core.pipeline.social_auth.load_extra_data',
+                'social_core.pipeline.user.user_details',
+                'users.pipeline.update_ad_groups',
+            ),
```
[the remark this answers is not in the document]
mikkokeskinen: I'd like it to be similar with the one used, but it doesn't really matter if they get out of sync. The main point is only the last part but to get there there needs to be all kinds of steps and some of the steps are not compatible with the social_core tests.

But I changed this to use the one from settings without the non-working steps.

auth_backends/tests/azure_ad_base.py
```diff
@@ -0,0 +1,171 @@
+import json
+from copy import copy
```
akikoskinen, line 2: Linter check says that `copy` is not used in this file.

akikoskinen (approved):

PR #243, opened 2022-06-14
attrs 21.4.0 version

kanguran:
https://www.attrs.org/en/stable/changelog.html


mikkokeskinen (changes requested):
Please use the same kind of commit messages as previously.

Like this:
`Update libsass from 0.14.5 to 0.21.0`

Also, it would be nice to know *why* the dependency was suddenly upgraded? Was there an error with the old version? Maybe a vulnerability?

PR #244, opened 2022-06-14
Fix Helsinki Azure AD backend backwards compatibility

mikkokeskinen:
The base class was changed from AzureADTenantOAuth2 to AzureADV2TenantOAuth2
in the previous change. I somehow missed that the value in "preferred_username"
claim was different from the "sub" claim used in the old backend.

That change meant that Social Auth would create a new user for the users that
had previously logged in with the old backend.

### DIFF
--- a/users/admin.py
+++ b/users/admin.py
@@ -20,9 +20,31 @@ from .models import LoginMethod, SessionElement, TunnistamoSession, User
 Application = get_application_model()
 
 
+class UserSocialAuthInline(admin.StackedInline):
+    model = UserSocialAuth
+    readonly_fields = ['provider', 'uid', 'get_extra_data']
+    exclude = ['extra_data']
+    extra = 0
+
+    def get_extra_data(self, obj=None):
+        if not obj or not obj.extra_data:
+            return ''
+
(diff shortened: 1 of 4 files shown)


users/admin.py
```diff
@@ -20,9 +20,31 @@
...
+    model = UserSocialAuth
+    readonly_fields = ['provider', 'uid', 'get_extra_data']
+    exclude = ['extra_data']
+    extra = 0
+
+    def get_extra_data(self, obj=None):
+        if not obj or not obj.extra_data:
+            return ''
+
+        return mark_safe(f'<pre>{json.dumps(obj.extra_data, indent=4)}</pre>')
```
akikoskinen, line 33: The result of `json.dumps` should probably be HTML escaped. Otherwise it could somehow be possible to inject some HTML, including javascript, into the page.

mikkokeskinen: Very true.

akikoskinen: Now it got fixed in the second commit. It could be correctly already in the first commit where that code is introduced.

akikoskinen (approved):

PR #265, opened 2023-02-08
Reauthentication support for social auth backends

akikoskinen:
Social login backends can be configured to perform reauthentication. Reauthentication happens when a user that has already previously authenticated (i.e. has a Django session) arrives again to the `/authorize` endpoint (with OIDC). Instead of directly returning the user back to the service where they came from, the user is redirected to the login endpoint, the same way as an unauthenticated user would be.

After this PR, the user is still requested to select a login method (if there are more than one available) in the login screen. This needs to be changed for the complete UX. This PR can still be merged, because if no social auth backends are configured to the new `ALWAYS_REAUTHENTICATE_BACKENDS` setting, then Tunnistamo works exactly the same as before this PR.

HP-1797


users/tests/conftest.py
```diff
@@ -305,14 +310,15 @@ def start_oidc_authorize(
         order=1,
     )
 
-    redirect_uris = ['https://example.com/callback']
-    oidc_client = oidcclient_factory(redirect_uris=redirect_uris)
+    redirect_uri = 'https://example.com/callback'
+    oidc_client_kwargs = {"redirect_uris": [redirect_uri]} | (oidc_client_kwargs or {})
```
mikkokeskinen, line 314: Or operator support in dicts is a Python 3.9 feature, I think. Could be nice to keep Tunnistamo compatible with (at least) all the supported Python 3 versions.

Even if the or would work with older Python versions I think this could be clearer as something like:
```python
    if oidc_client_kwargs is None:
        oidc_client_kwargs = {}
    oidc_client_kwargs.setdefault('redirect_uris', [redirect_uri])
```

oidc_apis/utils.py
```diff
@@ -71,6 +72,12 @@ def after_userlogin_hook(request, user, client):
     except OidcClientOptions.DoesNotExist:
         pass
 
+    is_returning_from_idp = request.GET.get('first_authz', '') == 'false'
+    if not is_returning_from_idp and last_login_backend in settings.ALWAYS_REAUTHENTICATE_BACKENDS:
+        next_page = request.get_full_path()
+        return get_authorize_endpoint_redirect_to_login_response(
+            next_page, oidc_provider.settings.get('OIDC_LOGIN_URL'))
+
```
mikkokeskinen, line 80: Could there be a comment describing why this is done? Maybe in the doc that already describes the other things the function is doing.

mikkokeskinen (changes requested):
Works nicely. Couple of comments.

Btw. commit message in 711bb30 has a typo. "distinguising" is missing an "h".

oidc_apis/utils.py
akikoskinen: I rewrote the function documentation and added this new feature to it too.

mikkokeskinen (approved):

PR #274, opened 2023-03-01
Redirect to social auth begin for reauthentication

akikoskinen:
When reauthentication is required for an already authenticated user, instead of redirecting to the login view and possibly giving the user a chance to select a login method, make the choice on behalf of the user (using the previously used login method) and go directly to the social auth flow.

I chose to do this this way instead of using the new `idp_hint` parameter, because the `idp_hint` parameter kind of indicates the original callers wish, which isn't the case in this scenario. Reusing the same parameter for this purpose could mess up things, because it gets stored in the session and may get used later. The solution proposed in this PR basically skips the login view, without a roundtrip to the browser.

### DIFF
--- a/users/views.py
+++ b/users/views.py
@@ -127,6 +127,10 @@ class LoginView(TemplateView):
         idp_hint = request.GET.get('idp_hint')
         login_methods = filter_login_methods_by_provider_ids_string(allowed_methods_for_client, idp_hint)
 
+        last_login_backend = request.session.get('social_auth_last_login_backend')
+        if last_login_backend in settings.ALWAYS_REAUTHENTICATE_BACKENDS:
+            login_methods = filter_login_methods_by_provider_ids_string(login_methods, last_login_backend)
+
         if login_methods is None:
             login_methods = LoginMethod.objects.all()
 
--- a/users/views.py
+++ b/users/views.py
@@ -85,6 +85,37 @@ def _get_allowed_login_methods_for_client_id(client_id):
     return allowed_methods
 
 
+def _generate_final_login_methods(login_methods, next_url, idp_hint):
+    methods = []
+
+    for login_method in login_methods:
+        if login_method.provider_id == 'saml':
+            continue  # SAML support removed
+
+        login_url_params = {}
+        if next_url:
+            login_url_params['next'] = next_url
+        if idp_hint:
+            login_url_params['idp_hint'] = idp_hint
+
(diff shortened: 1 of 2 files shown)


mikkokeskinen:
Until this point (as far as I know) the oidc provider and the social auth have been fully separate. This change would change that. I think it's a slippery slope.

In my opinion the `idp_hint` parameter would keep them separated and would work well. The `idp_hint` is saved to the session in social auth and used only if social auth redirects back to the login. Is there a concern I'm not seeing in that?

akikoskinen:
> Until this point (as far as I know) the oidc provider and the social auth have been fully separate. This change would change that. I think it's a slippery slope.

That's a good point.

> In my opinion the `idp_hint` parameter would keep them separated and would work well. The `idp_hint` is saved to the session in social auth and used only if social auth redirects back to the login. Is there a concern I'm not seeing in that?

Perhaps my main concern is the last few words in the original description of this PR and wasn't emphasized very clearly. Redirecting from the authorize endpoint to the login endpoint with the `idp_hint` makes a roundtrip to the browser. In other words that would be passing state via the browser, which isn't completely reliable. Meaning that it's possible that the `idp_hint` doesn't come back from the browser as it was – a rogue party can mess with it. If we really want to enforce something, such state and decisions should happen within the server.

I started thinking of another solution. What if the logic would be within the login view, which is already aware of social auth? It would check that

1. is the requester already logged in
1. with social auth
1. is the social auth backend in the `ALWAYS_REAUTHENTICATE_BACKENDS` setting

If the answer is yes to each of those, then it would make a redirect to the `social:begin` URL of that backend. What do you think?

mikkokeskinen:
Yeah, logic in the login view could work nicely.

akikoskinen:
Finally got back to this one. Now the envisioned logic is in the login view.

users/views.py
```diff
@@ -85,6 +85,37 @@ def _get_allowed_login_methods_for_client_id(client_id):
     return allowed_methods
 
 
+def _generate_final_login_methods(login_methods, next_url, idp_hint):
```
mikkokeskinen, line 88: Could something like `_generate_login_url_parameter_to_login_methods` be a better name for this function?

akikoskinen: The function does also some filtering, possibly dropping some login methods away. That's why I came up with the peculiar name.

mikkokeskinen (approved):

PR #276, opened 2023-03-09
Consider idp_hint parameter in authorize endpoint

akikoskinen:
When a previously authenticated user authenticates using another OIDC client, it's checked that the login method that the user used before, is good also for this new client. If not, the user gets logged out first.

Now the login methods of the new OIDC client are first filtered with the `idp_hint` parameter, and only after that the check is made.


mikkokeskinen (approved):
👍

PR #282, opened 2023-05-17
Duplicate social login fix

mikkokeskinen:
When the user was logged in with a backend that is in the
ALWAYS_REAUTHENTICATE_BACKENDS setting, there could be a case where
another log in with the same backend but with a different user would add
a second social auth to the user.

It happened if the user was still logged in to Tunnistamo but logged
out of the backend.

Refs HP-1934

### DIFF
--- a/users/tests/test_pipeline.py
+++ b/users/tests/test_pipeline.py
@@ -1,0 +1,55 @@
+import pytest
+from django.contrib import auth
+from django.urls import reverse
+from django.utils.crypto import get_random_string
+from social_core.exceptions import AuthFailed
+
+from tunnistamo.tests.conftest import create_rsa_key, reload_social_django_utils
+from users.tests.conftest import DummyFixedOidcBackend
+
+
+@pytest.fixture
+def dummy_backend(settings):
+    create_rsa_key()
+
+    settings.AUTHENTICATION_BACKENDS = settings.AUTHENTICATION_BACKENDS + (
+        'users.tests.conftest.DummyFixedOidcBackend',
+    )
+    settings.SOCIAL_AUTH_DUMMYFIXEDOIDCBACKEND_OIDC_ENDPOINT = 'https://dummy.example.com'
+    settings.SOCIAL_AUTH_DUMMYFIXEDOIDCBACKEND_KEY = 'tunnistamo'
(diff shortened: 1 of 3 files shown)


users/tests/test_pipeline.py
```diff
@@ -0,0 +1,54 @@
...
+def dummy_backend(settings):
+    create_rsa_key()
+
+    settings.AUTHENTICATION_BACKENDS = settings.AUTHENTICATION_BACKENDS + (
+        'users.tests.conftest.DummyFixedOidcBackend',
+    )
+    settings.SOCIAL_AUTH_DUMMYFIXEDOIDCBACKEND_OIDC_ENDPOINT = 'https://dummy.example.com'
+    settings.SOCIAL_AUTH_DUMMYFIXEDOIDCBACKEND_KEY = 'tunnistamo'
+    settings.EMAIL_EXEMPT_AUTH_BACKENDS = [DummyFixedOidcBackend.name]
+    settings.ALWAYS_REAUTHENTICATE_BACKENDS = [DummyFixedOidcBackend.name]
```
akikoskinen, line 20: Although `ALWAYS_REAUTHENTICATE_BACKENDS` revealed the bug, it's not the cause of it. This setting doesn't have to be present in these tests. I would remove it, otherwise it could cause confusion why it's here.

users/tests/test_pipeline.py
```diff
@@ -0,0 +1,54 @@
...
+    # Change the sub value and complete dummy backend auth again
+    settings.SOCIAL_AUTH_DUMMYFIXEDOIDCBACKEND_SUB_VALUE = 'second_sub_value'
+
+    # The social login should fail because it shouldn't be possible to
+    # attach a second social login to the already logged-in user with the same backend
+    # but with a different user in the backend.
+    with pytest.raises(AuthFailed):
+        request_social_complete(client)
+
+    assert user.social_auth.count() == 1
```
akikoskinen, line 54: This test could also check that the user has been logged out.

users/tests/test_pipeline.py
mikkokeskinen: Ah. True.

users/tests/test_pipeline.py
mikkokeskinen: Done.

akikoskinen (approved):

PR #291, opened 2023-08-03
Dockerfile: use netcat-openbsd

charn:
Base image uses Debian bookworm which doesn't have the netcat package.


akikoskinen (approved):
:+1:

mikkokeskinen (approved):

PR #293, opened 2023-08-16
Remove allauth from installed apps and requirements

mikkokeskinen:


requirements.txt
```diff
@@ -134,19 +130,16 @@ pyjwkest==1.4.0
     # via
     #   django-oidc-provider
     #   drf-oidc-auth
-pyjwt[crypto]==2.4.0
+pyjwt==2.4.0
```
akikoskinen, line 133: Would it beneficial to keep this requirement as `pyjwt[crypto]`? I guess that could be achieved by changing the line (or lines? there's both `pyjwt` and `PyJWT` :thinking:) in `requirements.in`.

There's some reasoning in PyJWT's [installation guide](https://pyjwt.readthedocs.io/en/stable/installation.html#cryptographic-dependencies-optional) in favour of marking the dependency to `cryptography` explicitly.

mikkokeskinen: Yeah, I guess it's good to be explicit. Changed the requirements.in a bit.

akikoskinen (approved):

PR #306, opened 2024-02-22
HP-2219 | Users can use helsinki_tunnus heltunnistussuomifi simultaneously

charn:
It was not possible for a linked keycloak account (i.e. an account which can use weak or strong authentication) to log into Helsinki profile.


voneiden (approved):
seems good :pray:

PR #308, opened 2024-03-18
HP-2279 | feat: try to associate user by keycloak UUID

charn:
User might have a social account with another Keycloak provider. Some fields between Tunnistamo and Keycloak might not be in sync (e.g. email), so we'll use UUID to attempt to find the user.


voneiden (approved):
Seems pretty good

PR #309, opened 2024-05-13
fix: increase helsinki azure AD group listing size

voneiden:
Microsoft Graph memberOf API defaults to page size of 100. Increase it to 999 like it is in helsinki-tunnistus-plugin as we have users who have 100+ groups.

Refs: HP-2432


charn (approved):
LGTM! :+1: :shipit:

PR #310, opened 2024-05-24
feat: support Entra guest user groups

voneiden:
Entra guest users have limited visibility into groups via the API. Basically they can only see the group id's they belong to. Both displayName and securityEnabled values are null. However filtering in the request using `$filter=securityEnabled eq true` still works for guest users. This feature allows returning group UUIDs for guest users.

Refs: HP-2434

### DIFF
--- a/auth_backends/helsinki_azure_ad.py
+++ b/auth_backends/helsinki_azure_ad.py
@@ -18,7 +18,13 @@ class HelsinkiAzureADTenantOAuth2(AzureADV2TenantOAuth2):
     name_logged_in_to = pgettext('logged in to []', 'Azure AD')
     name_logout_from = pgettext('log out from []', 'Azure AD')
     name_goto = pgettext('go to []', 'Azure AD')
-    USER_DATA_URL = 'https://graph.microsoft.com/v1.0/me/memberOf?$select=id,displayName,securityEnabled&$top=999'
+    USER_DATA_URL = (
+        'https://graph.microsoft.com/v1.0/me/memberOf?'
+        '$filter=securityEnabled eq true'
+        '&$count=true'
+        '&$top=999'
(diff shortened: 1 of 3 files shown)


auth_backends/helsinki_azure_ad.py
```diff
@@ -18,7 +18,11 @@ class HelsinkiAzureADTenantOAuth2(AzureADV2TenantOAuth2):
     name_logged_in_to = pgettext('logged in to []', 'Azure AD')
     name_logout_from = pgettext('log out from []', 'Azure AD')
     name_goto = pgettext('go to []', 'Azure AD')
-    USER_DATA_URL = 'https://graph.microsoft.com/v1.0/me/memberOf?$select=id,displayName,securityEnabled'
+    USER_DATA_URL = (
+            'https://graph.microsoft.com/v1.0/me/memberOf?'
+            '$select=id,displayName,securityEnabled'
+            '$filter=securityEnabled eq true'
```
charn, line 24: Is this missing `&`? Also [Microsoft Graph Explorer ](https://developer.microsoft.com/en-us/graph/graph-explorer) says the following when adding the `filter` part to the query `The specified filter to the reference property query is currently not supported.`

```suggestion
            '&$filter=securityEnabled eq true'
```

charn:
Refactored the Microsoft API calls into an "advanced query" which allows filtering with `securityEnabled`.

Filtering groups by the `securityEnabled` flag requires using an "advanced query", which requires adding `$count=true` query parameter and a `ConsistencyLevel: eventual` HTTP header. :shrug: 

Also we'll request the maximum possible number of groups. By default the groups query returns a maximum of 100 results. It can return a maximum of 999 groups.

tuomas777 (dismissed):
LGTM 👍 🫏

PR #313, opened 2024-07-03
HP-2461 | feat: pass keycloak action parameter through

charn:
Running actions with keycloak is done using `kc_action` parameter. Here, we're just passing the action through so that a user can e.g. update the password on keycloak while the authentication still flows through tunnistamo.

### DIFF
--- a/users/views.py
+++ b/users/views.py
@@ -166,9 +166,9 @@ class TunnistamoOidcAuthorizeView(AuthorizeView):
     def get(self, request, *args, **kwargs):
         request.GET = _extend_scope_in_query_params(request.GET)
 
-        if request.GET.get('client_id'):
+        if client_id := request.GET.get("client_id"):
             try:
-                client = Client.objects.get(client_id=request.GET.get('client_id'))
+                client = Client.objects.get(client_id=client_id)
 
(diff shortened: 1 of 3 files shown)


users/views.py
```diff
@@ -178,6 +178,11 @@ def get(self, request, *args, **kwargs):
                 # validated again in the parent get method.
                 pass
 
+        # Keycloak action passthrough
+        if request.GET.get('kc_action'):
+            print(request.GET.get('kc_action'))
```
nicobav, line 183: is this some leftover from debugging? Could be removed or changed to be outputted via logger.

users/views.py
```diff
@@ -178,6 +178,11 @@ def get(self, request, *args, **kwargs):
                 # validated again in the parent get method.
                 pass
 
+        # Keycloak action passthrough
+        if request.GET.get('kc_action'):
```
nicobav, line 182: This could be walrusified := 
(double quotes would be nice in; `.get("kc_action")`

nicobav (changes requested):

users/views.py
charn: Did some refactoring. :+1:

nicobav (approved):
LGTM :shipit:

PR #314, opened 2024-08-07
HP-2491 | return kc_action_status from Keycloak on authentication

charn:
`kc_action_status` is used to indicate e.g. whether a requested password change was successful or not. It will be returned
from tunnistamo along with `code` and `state` to the callback url the same way keycloak does it. This should make a later
transtition to keycloak (by dropping tunnistamo from the middle) not require extra work for any logic using `kc_action_status`.

Upon login the `kc_action_status` is saved to the session from which it is then retrieved in `TunnistamoAuthorizeEndpoint`.

### DIFF
--- a/users/tests/test_pipeline.py
+++ b/users/tests/test_pipeline.py
@@ -1,5 +1,8 @@
+from urllib.parse import urlencode
+
 import pytest
 from django.contrib import auth
+from django.contrib.sessions.middleware import SessionMiddleware
 from django.urls import reverse
 from django.utils.crypto import get_random_string
 from social_core.exceptions import AuthFailed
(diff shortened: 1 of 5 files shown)


users/tests/test_pipeline.py
```diff
@@ -103,3 +106,35 @@ def test_association_by_keycloak_uuid(login_backend, existing_backend, user, exp
...
         assert pipeline_data is None
+
+
+@pytest.mark.parametrize(
+    "backend_name,expected",
+    (
+        ("heltunnistussuomifi", True),
+        ("helsinki_tunnus", True),
+        ("helsinki_adfs", False),
+    ),
```
danipran, line 117: If I see something called `expected` in unit tests, I expect it to be the expected value. Instead, this is a flag for changing the expected value later in the test. Just change this to the actual expected values.

If you're worried about `None == None` (which should be fine in this case IMO), just set a default value in `session_get`, e.g. `session_get("kc_action_status", "NOT SET")`. Or at least I imagine it should have a default option. Then you can definitely just assert for `strategy.blabla == expected`.

charn: Refactored. :+1:

danipran (approved):
LGTM! 👍

PR #315, opened 2024-09-24
HP-2655 | fix: uwsgi has version mismatches with precompiled python libs

charn:
Instead of using libxml2 we can use expat with uwsgi which avoids the version mismatch with precompiled python libraries.

Also ssl is not terminated by uwsgi, so the feature can be disabled.

Refs: HP-2655


voneiden (dismissed):
Seems interesting.

